PR #5, opened 2018-06-13
implement in-memory db and fix tests

jeremydeanlakey:


BlueDBInMemory/src/main/java/io/bluedb/memory/BlueCollectionImpl.java
```diff
@@ -1,71 +1,107 @@
...
 import io.bluedb.api.Updater;
 import io.bluedb.api.exceptions.BlueDbException;
 import io.bluedb.api.keys.BlueKey;
+import io.bluedb.api.keys.TimeFrameKey;
+import io.bluedb.api.keys.TimeKey;
 
 class BlueCollectionImpl<T extends Serializable> implements BlueCollection<T> {
 
 	private Class<T> type;
+	private Map<BlueKey, T> data = new HashMap<>();
```
d-w-johnson: Should we use a concurrent hash map? Or does this really matter for the in memory impl?

BlueDBInMemory/src/main/java/io/bluedb/memory/BlueDbInMemory.java
```diff
@@ -1,21 +1,29 @@
...
+import java.util.HashMap;
+import java.util.Map;
 import io.bluedb.api.BlueDb;
 import io.bluedb.api.exceptions.BlueDbException;
 import io.bluedb.api.BlueCollection;
 
 public class BlueDbInMemory implements BlueDb {
 
+	@SuppressWarnings("rawtypes")
+	Map<Class, BlueCollection> collections = new HashMap<>();
```
d-w-johnson: You could declare the key type to be Class<?> instead of ignoring the warning. Or you could do Class<? extends Serializable>. Doesn't really matter much.

BlueDBInMemory/src/main/java/io/bluedb/memory/BlueDbInMemory.java
```diff
@@ -1,21 +1,29 @@
...
+
+	@SuppressWarnings("unchecked")
 	@Override
 	public <T extends Serializable> BlueCollection<T> getCollection(Class<T> type) {
-		return new BlueCollectionImpl<>(type);
+		if (!collections.containsKey(type)) {
```
d-w-johnson: We don't really save much from only having one instance of each collections right? We just don't have to construct them every time? Having a client use the same collection on different threads isn't a problem right? I don't think it is. Technically this getCollection method could result in multiple collections getting created if its called at the same time from multiple threads. Its really not a problem to create more than one though.

BlueDBInMemory/src/main/java/io/bluedb/memory/BlueCollectionImpl.java
```diff
@@ -1,71 +1,107 @@
...
-	public List<T> getList() {
-		// TODO Auto-generated method stub
-		return null;
+	public List<T> getList(long minTime, long maxTime, List<Condition<T>> objectConditions) {
+		List<T> results = new ArrayList<>();
+		List<BlueKey> matches = findMatches(minTime, maxTime, objectConditions);
+		for (BlueKey key: matches) {
+			results.add(data.get(key));
+		}
+		return results;
```
d-w-johnson: Order this by time if they have a start time

BlueDBInMemory/src/main/java/io/bluedb/memory/BlueCollectionImpl.java
```diff
@@ -107,6 +109,7 @@ public void updateAll(long minTime, long maxTime, List<Condition<T>> objectCondi
 				results.add(key);
 			}
 		}
+		sort(results);
```
d-w-johnson: Is it important to delete and update in order? Doesn't really matter for the in memory impl, but might on disk.

PR #7, opened 2018-06-19
Prototype

jeremydeanlakey:


BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/SegmentIdConverter.java
```diff
@@ -0,0 +1,27 @@
...
+	public static long convertTimeToSegmentId(long time) {
+		return time / MILLIS_IN_BLOCK;
+	}
+
+	public static long convertSegmentIdToTime(long segmentId) {
+		return segmentId * MILLIS_IN_BLOCK;
+	}
+
+	public static List<Long> getSegments(long min, long max) {
+		List<Long> segments = new ArrayList<>();
```
d-w-johnson: You could just calculate the start block and the end block and you know everything in between will be included as well. Reduces the number of calculations

PR #11, opened 2018-06-25
Segment manager

jeremydeanlakey:
The main point of this pull request is to implement SegmentManager, whose sole purpose is to determine which segments are relevant to a query and provide them.  The logic to do this was in the BlueCollection class but was getting too complicated once we started implementing a more complicated file structure.


BlueDBOnDisk/src/main/java/io/bluedb/disk/recovery/RecoveryManager.java
```diff
@@ -26,12 +27,14 @@ public RecoveryManager(BlueCollectionImpl<T> collection) {
...
 		String filename = getFileName(change);
-		Blutils.save(filename, change, collection.getSerializer());
+		Path path = Paths.get(".", ".pending", filename);
+		Blutils.save(path.toString(), change, collection.getSerializer());
 	}
 
 	public void removeChange(PendingChange<T> change) throws BlueDbException {
 		String filename = getFileName(change);
-		File file = new File(filename);
+		Path path = Paths.get(".", ".pending", filename);
```
d-w-johnson: This should reference the path member that should be set in the constructor

BlueDBOnDisk/src/main/java/io/bluedb/disk/Blutils.java
```diff
@@ -18,6 +19,7 @@
 public class Blutils {
 	public static void save(String path, Object o, BlueSerializer serializer) throws BlueDbException {
```
d-w-johnson: Lets pass in a Path object here instead of a String

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -7,88 +7,129 @@
...
-
-	public Segment(Path collectionPath, long segmentId, BlueSerializer serializer) {
-		this.path = Paths.get(collectionPath.toString(), segmentId + SUFFIX);
-		this.pathString = this.path.toString();
+	public Segment(Path segmentPath, BlueSerializer serializer) {
+		this.path = segmentPath;
+//		this.pathString = this.path.toString();
 		this.serializer = serializer;
 	}
+//
```
d-w-johnson: We can just delete this stuff that is now commented out

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -7,88 +7,129 @@
...
-		TreeMap<BlueKey, BlueEntity<T>> data = load();
+		TreeMap<BlueKey, BlueEntity<T>> data = load(key.getGroupingNumber());
 		data.remove(key);
+		String pathString = Paths.get(path.toString(), String.valueOf(key.getGroupingNumber())).toString();
 		Blutils.save(pathString, data, serializer);
 	}
 
 	public BlueEntity<T> read(BlueKey key) throws BlueDbException {
-		Collection<BlueEntity<T>> values = load().values();
+		Collection<BlueEntity<T>> values = load(key.getGroupingNumber()).values();
```
d-w-johnson: The load gives you a map. You should just be able to call get and pass in the key instead of iterating through all of the values.

Though, we talked about getting rid of the TreeMap for starters.

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -7,88 +7,129 @@
...
-
 	public List<BlueEntity<T>> read(long minTime, long maxTime) throws BlueDbException {
 		List<BlueEntity<T>> results = new ArrayList<>();
-		for (BlueEntity<T>entity: load().values()) {
+		List<Long> groupingNumbers = getAllFiles().stream()
+				.filter((f) -> isFileNameALong(f))
+				.map((f) -> f.getName())
+				.map((filename) -> Long.valueOf(filename))
+				.collect(Collectors.toList());
+		for (long groupingNumber: groupingNumbers) {
```
d-w-johnson: You could choose to do a forEach here instead of collecting then iterating over the list again. A little more efficient though less clean.

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -7,88 +7,129 @@
...
+	private TreeMap<BlueKey, BlueEntity<T>> load(long groupingNumber) throws BlueDbException {
+		Path groupPath = Paths.get(path.toString(), String.valueOf(groupingNumber));
 		try {
-			File file = new File(pathString);
+			File file = groupPath.toFile();
 			if (!file.exists()) {
 				return new TreeMap<>();
 			} else {
-				byte[] bytes = Files.readAllBytes(Paths.get(pathString));
+				byte[] bytes = Files.readAllBytes(groupPath);
```
d-w-johnson: We could probably use another method in serializer called "readObjectFromFile" that everything can easily use

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -1,94 +1,201 @@
...
+	protected Path getPath() {
+		return segmentPath;
+	}
+
+	private Path getPathFor(BlueKey key) {
+		String fileName = String.valueOf(key.getGroupingNumber());
+		return Paths.get(segmentPath.toString(), fileName);
+	}
+
+	private TreeMap<BlueKey, T> fetchAsMap(File file) throws BlueDbException {
```
d-w-johnson: You have to iterate through every item in the list to convert it to a map and every item in the map to convert it back to a list. Can't we just iterate through the list one time to identify and update/delete the item(s) that we care about? That would save a lot of cycles and memory. I don't see what the map gives us that isn't pretty easy to do on our own.

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
jeremydeanlakey: I'd like to punt this to FileManager, which is probably what I'll implement next.

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/SegmentManager.java
```diff
@@ -0,0 +1,164 @@
...
+		long groupingNumber = key.getGroupingNumber();
+		return getPath(groupingNumber);
+	}
+
+	protected Path getPath(long groupingNumber) {
+		String level0 = getRangeFileName(groupingNumber, LEVEL_0);
+		String level1 = getRangeFileName(groupingNumber, LEVEL_1);
+		String level2 = getRangeFileName(groupingNumber, LEVEL_2);
+		String level3 = getRangeFileName(groupingNumber, LEVEL_3);
+		return Paths.get(collection.getPath().toString(), level0, level1, level2, level3 + SEGMENT_SUFFIX);
```
d-w-johnson: I'm still a little confused how this works. Maybe we could white board it or something tomorrow. I think looking at it in eclipse will be a bit easier.

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -151,6 +128,30 @@ private static boolean inTimeRange(long minTime, long maxTime, BlueKey key) {
 			return key.getGroupingNumber() >= minTime && key.getGroupingNumber() <= maxTime;
 		}
 	}
+	
+	protected static <T extends Serializable> T remove(BlueKey key, List<BlueEntity<T>> entities) {
```
d-w-johnson: Usually when I do something like this I just us an iterator so that I can call .remove on it if it matches.

Since this is an array list it is fine to call get(i) and remove(i), but if you're doing this way I would change the entities parameter to be an ArrayList to ensure that nobody calls this method with a LinkedList or something that would be incredibly inefficient.
I would change the entities parameter to an ArrayList if we're going to call get(i) in the method. I know you are passing in an ArrayList but this enforces

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -37,38 +34,41 @@ public boolean contains(BlueKey key) throws BlueDbException {
...
 	}
 
 	public void put(BlueKey key, T value) throws BlueDbException {
 		File file = getFileFor(key);
-		Map<BlueKey, T> fileContents = fetchAsMap(file);
-		fileContents.put(key, value);
-		persist(file, fileContents);
+		ArrayList<BlueEntity<T>> entities = fetch(file);
+		BlueEntity<T> newEntity = new BlueEntity<T>(key, value);
+		remove(key, entities);
```
d-w-johnson: Should we be throwing an exception if the key already exists?

PR #15, opened 2018-06-27
Serializers by collection and one collection per class/name pair

d-w-johnson:


BlueDBOnDisk/src/main/java/io/bluedb/disk/BlueDbOnDisk.java
```diff
@@ -2,28 +2,36 @@
...
 	}
 
 	@Override
-	public <T extends Serializable> BlueCollection<T> getCollection(Class<T> type, String name) {
-		// TODO make sure only one Collection object per collection to avoid concurrency issues
-		return new BlueCollectionImpl<>(this, type);
+	@SuppressWarnings("unchecked")
+	public <T extends Serializable> BlueCollection<T> getCollection(Class<T> type, String name) throws BlueDbException {
+		synchronized (collections) {
+			String key = type.getName() + "-" + name;
```
d-w-johnson: I made the type and name be the key of the collection. This means that if they give us the same name for different types it will still work. This isn't really what we discussed but is there a problem with it? I guess there would be a collision on the directories... I'll change this.

PR #20, opened 2018-07-02
Locks and streams

jeremydeanlakey:
We need to store data in a way that allows for streaming reads with FST.  I've implemented a couple classes to facilitate writing and reading data that way.


jeremydeanlakey:
Closing because we might want to take a different route.

PR #24, opened 2018-07-04
streaming

jeremydeanlakey:


BlueDBOnDisk/src/main/java/io/bluedb/disk/file/BlueObjectInput.java
```diff
@@ -0,0 +1,64 @@
...
+		this.path = readLock.getKey();
+		this.serializer = serializer;
+		dataInputStream = openDataInputStream(path.toFile());
+	}
+
+	public T next() throws EOFException, BlueDbException {
+		int objectLength;
+		try {
+			objectLength = dataInputStream.readInt();
+			byte[] buffer = new byte[objectLength];
```
d-w-johnson: It would be nice if we can figure out a good way to re-use our buffer

BlueDBOnDisk/src/main/java/io/bluedb/disk/file/BlueObjectInput.java
```diff
@@ -0,0 +1,64 @@
...
+			int bytesRead = dataInputStream.read(buffer,0, objectLength);
```
d-w-johnson: I feel like it may not read the entire length which is why their reference code that I copied keeps reading until the amount of bytes have been read or the end of the file is reached.

BlueDBOnDisk/src/main/java/io/bluedb/disk/file/BlueObjectInput.java
jeremydeanlakey: There's probably a way but I don't want to spend a lot of time on it until we determine if we're for sure going to go this route

BlueDBOnDisk/src/main/java/io/bluedb/disk/file/BlueObjectInput.java
jeremydeanlakey: I'm switching to readFully, which should block until the bytes are available.  Let me know if you think that way is problematic.

PR #26, opened 2018-07-05
Rollup

jeremydeanlakey:
This is just a really rough initial implementation.  We'll likely want to do a lot of fine tuning.

### DIFF
--- a/BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/RollupScheduler.java
+++ b/BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/RollupScheduler.java
@@ -5,13 +5,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
-import io.bluedb.api.exceptions.BlueDbException;
 import io.bluedb.disk.collection.BlueCollectionImpl;
 
 public class RollupScheduler {
 
-	private static final long WAIT_BETWEEN_REVIEWS = 3_000; // TODO something more sophisticated than just wait ?
-	private static final long WAIT_BEFORE_ROLLUP = 3_000; // TODO 
+	private static final long WAIT_BETWEEN_REVIEWS = 30_000; // TODO something more sophisticated than just wait ?
+	private static final long WAIT_BEFORE_ROLLUP = 3600_000; // TODO something more sophisticated?
 
 	private final BlueCollectionImpl<?> collection;
 	private final Map<TimeRange, Long> lastInsertTimes;
@@ -28,7 +27,6 @@ public class RollupScheduler {
 	}
 
 	public void reportInsert(TimeRange timeRange) {
-		System.out.println("RollupScheduler.reportInsert(" + timeRange + ")");
 		lastInsertTimes.put(timeRange, System.currentTimeMillis());
 	}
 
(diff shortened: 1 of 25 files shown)


BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -116,6 +116,47 @@ public T get(BlueKey key) throws BlueDbException {
...
+		List<File> filesToRollup = getFilesInRange(start, end);
+		Path path = Paths.get(segmentPath.toString(), start + "_" + end);
+		Path tmpPath = FileManager.createTempFilePath(path);
+		List<BlueEntity<T>> entities = new ArrayList<>();
+
+		// TODO switch to streaming
+		// TODO keep sorted, here and everywhere you write
+		for (File file: filesToRollup) {
+			List<BlueEntity<T>> es = fetch(file);
+			entities.addAll(es);
```
d-w-johnson: I'd rather not have to hold all entities being rolled up in memory at one time. Instead, we could stream through each file one at a time while writing each object to the new temp file. Holding a write lock for awhile on that temp file shouldn't be an issue.

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -116,6 +116,47 @@ public T get(BlueKey key) throws BlueDbException {
...
+
+		List<BlueWriteLock<Path>> sourceFileWriteLocks = new ArrayList<>();
+		try (BlueWriteLock<Path> targetFileLock = lockManager.acquireWriteLock(path)){
+			for (File file: filesToRollup) {
+				sourceFileWriteLocks.add(lockManager.acquireWriteLock(file.toPath()));
+			}
+
+			FileManager.moveFile(tmpPath, targetFileLock);
+			for (BlueWriteLock<Path> writeLock: sourceFileWriteLocks) {
+				FileManager.deleteFile(writeLock);
```
d-w-johnson: Could something be added to one of the files being rolled up after we've already added all of its entities to our list? If so, they didn't make it into the temp file and we'll lose them.

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/RollupScheduler.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.concurrent.ConcurrentHashMap;
+import io.bluedb.api.exceptions.BlueDbException;
+import io.bluedb.disk.collection.BlueCollectionImpl;
+
+public class RollupScheduler {
+
+	private static final long WAIT_BETWEEN_REVIEWS = 3_000; // TODO something more sophisticated than just wait ?
+	private static final long WAIT_BEFORE_ROLLUP = 3_000; // TODO
```
d-w-johnson: This should start much higher right?

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
jeremydeanlakey: using streaming in the latest updates

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
jeremydeanlakey: not right now.  the rollup is currently done in the write thread.  If we change that, we'll have to change a bunch of logic.

PR #29, opened 2018-07-07
Streaming processing

jeremydeanlakey:
This is mostly about reading and writing in a streaming way, although there is some other improvements thrown in. Changes also reflects the other pending pull request.

### DIFF
--- a/BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
+++ b/BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
@@ -100,10 +100,10 @@ public class Segment <T extends Serializable> {
 
     public List<BlueEntity<T>> getRange(long minTime, long maxTime) throws BlueDbException {
 		List<BlueEntity<T>> results = new ArrayList<>();
-		List<File> filesInFolder = FileManager.getFolderContents(segmentPath.toFile());
+		File folder = segmentPath.toFile();
 		// Note that we cannot bound this from below because a TimeRangeKey that overlaps the target range
 		//      will be stored at the start time;
-		List<File> relevantFiles = Blutils.filter(filesInFolder, (f) -> doesfileNameRangeOverlap(f, Long.MIN_VALUE, maxTime));
+		List<File> relevantFiles = FileManager.getFolderContents(folder, (f) -> doesfileNameRangeOverlap(f, Long.MIN_VALUE, maxTime));
 		for (File file: relevantFiles) {
 			List<BlueEntity<T>> fileContents = fetch(file);
 			for (BlueEntity<T> entity: fileContents) {
--- a/BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
+++ b/BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
@@ -7,6 +7,7 @@ import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
 import io.bluedb.api.exceptions.BlueDbException;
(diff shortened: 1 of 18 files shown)


BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -40,41 +41,104 @@ public boolean contains(BlueKey key) throws BlueDbException {
...
+			@Override
+			public void process(BlueObjectInput<BlueEntity<T>> input, BlueObjectOutput<BlueEntity<T>> output) throws BlueDbException {
+				BlueEntity<T> newEntity = new BlueEntity<T>(newKey, newValue);
+				while (input.hasNext()) {
+					BlueEntity<T> iterEntity = input.next();
+					BlueKey iterKey = iterEntity.getKey();
+					if (iterKey.equals(newKey)) {
+						output.write(newEntity);
+						newEntity = null;
+					} else if (newEntity != null && iterKey.getGroupingNumber() > groupingNumber) {
```
d-w-johnson: So if I delete something then try to update it then the object will be resurrected? Or I guess in order to have the newValue that means that we already found the value once and applied the update lambda to it right? Why do we end up needing this branch?

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -40,41 +41,104 @@ public boolean contains(BlueKey key) throws BlueDbException {
...
-		try (BlueReadLock<Path> readLock = getReadLockFor(key)) {
-			if (readLock != null) {
-				entities = fetch(readLock);
-				file = readLock.getKey().toFile();
-			} else {
-				entities = new ArrayList<>();
-				file = 	getPathFor(key, 1).toFile();
+	public void update(BlueKey newKey, T newValue) throws BlueDbException {
+		long groupingNumber = newKey.getGroupingNumber();
+		modifyChunk(groupingNumber, new Processor<T>() {
```
d-w-johnson: I don't like creating a new anonymous inner class every time for all of these processes. I'd extract the contents into a class or method and use a lambda or method reference. Or at lease create a private `updateProcessor` member that is an anonymous inner class and can be used every time.

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -41,6 +43,10 @@ public boolean contains(BlueKey key) throws BlueDbException {
 		return get(key) != null;
 	}
 
+	interface Processor<X extends Serializable> {
```
d-w-johnson: I also think this deserves its own class

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -28,10 +28,19 @@
...
+	private final LockManager<Path> lockManager;
 
 	public Segment(Path segmentPath, FileManager fileManager) {
 		this.segmentPath = segmentPath;
 		this.fileManager = fileManager;
+		lockManager = fileManager.getLockManager();
+	}
+
+	// for testing only
+	protected Segment() {
```
d-w-johnson: How do you feel about making a static method called `createTestSegment()` or something like that? Then you'd easily be able to make a test instance of the class without the risk of someone trying to use it in non-test environments.

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
jeremydeanlakey: removed recreate on update

PR #31, opened 2018-07-11
iterators

jeremydeanlakey:


BlueDBOnDisk/src/main/java/io/bluedb/disk/file/BlueObjectOutput.java
```diff
@@ -28,6 +29,8 @@ public BlueObjectOutput(BlueWriteLock<Path> writeLock, BlueSerializer serializer
 
 	// for testing only
 	protected BlueObjectOutput(Path path, BlueSerializer serializer, DataOutputStream dataOutputStream) {
+		LockManager<Path> lockManager = new LockManager<Path>();
```
d-w-johnson: Same question about the static createTest method

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -168,25 +159,21 @@ public T get(BlueKey key) throws BlueDbException {
 		long minGroupingNumber = Long.MIN_VALUE;
 		TimeRange timeRange = new TimeRange(Long.MIN_VALUE, max);
 		List<File> relevantFiles = getOrderedFilesInRange(timeRange);
+		List<TimeRange> timeRanges = relevantFiles.stream().map((f) -> TimeRange.fromUnderscoreDelmimitedString(f.getName())).collect(Collectors.toList());
```
d-w-johnson: Could you split this onto multiple lines?

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -155,30 +155,38 @@ public T get(BlueKey key) throws BlueDbException {
...
-						continue;
-					}
-					if (Blutils.isInRange(key, min, max))
-						results.add(next);
-				}
-			}
-			minGroupingNumber = rangeForThisFile.getEnd() + 1;
 		}
 		return results;
+
```
d-w-johnson: If this commented code isn't useful anymore can we get rid of it?

BlueDB/src/main/java/io/bluedb/api/CloseableIterator.java
```diff
@@ -0,0 +1,7 @@
+package io.bluedb.api;
+
+import java.io.Closeable;
+import java.util.Iterator;
+
+public interface CloseableIterator<E> extends Closeable, Iterator<E> {
```
d-w-johnson: How do you decide what letter to use in your generic templates? For example `E` or `X` instead of `T`. I've always just used `T`. I'm just curious.

BlueDB/src/main/java/io/bluedb/api/BlueQuery.java
```diff
@@ -16,7 +15,7 @@
 	BlueQuery<T> afterOrAtTime(long time);
 
 	List<T> getList() throws BlueDbException;
-	Iterator<T> getIterator() throws BlueDbException;
+	CloseableIterator<T> getIterator() throws BlueDbException;
```
d-w-johnson: I would prefer returning a class with stream in the name so that it screams "This is a resource and must be used in a try with resource!". Iterator doesn't scream that at me. Either that or make the iterator name contain resource or something. Then depending on what the class is we could change the method name to match.

BlueDB/src/main/java/io/bluedb/api/CloseableIterator.java
jeremydeanlakey: Eclipse automatically populated Iterator with E (presumable Element?) so I went with it.  Mostly I use T or for a generic static factory method, I use X.

PR #34, opened 2018-07-13
New paths

jeremydeanlakey:


BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -28,7 +28,7 @@
 
 public class Segment <T extends Serializable> {
 
-	private final static Long SEGMENT_SIZE = SegmentManager.LEVEL_3;
+	private final static Long SEGMENT_SIZE = SegmentManager.SIZE_SEGMENT;
```
d-w-johnson: Segment sizes are going to be bigger for timeless segments right? So we'll need this to be dynamic if we are able to use the same segment class for time and timeless segments.

PR #37, opened 2018-07-16
Rollup recovery

jeremydeanlakey:
this should be done after max values and timeout.

### DIFF
--- a/BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
+++ b/BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
@@ -13,6 +13,7 @@ import java.util.List;
 
 import io.bluedb.api.exceptions.BlueDbException;
 import io.bluedb.api.keys.BlueKey;
+import io.bluedb.disk.Blutils;
 import io.bluedb.disk.file.BlueObjectInput;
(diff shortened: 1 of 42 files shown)


BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -200,11 +199,8 @@ protected static boolean doesfileNameRangeOverlap(File file, long min, long max
 		for (long rollupLevel: ROLLUP_LEVELS) {
 			Path path = getPathFor(groupingNumber, rollupLevel);
 			BlueReadLock<Path> lock = lockManager.acquireReadLock(path);
-			try {
-				if (lock.getKey().toFile().exists()) {
-					return lock;
-				}
-			} catch (Throwable t) { // make sure we don't hold onto the lock if there's an exception
+			if (lock.getKey().toFile().exists()) {
```
d-w-johnson: Is it not possible to throw an exception here then? Seems scary to not be closing the lock in a finally

jeremydeanlakey: I think we can add a try-catch block that only closes the lock if we have an exception

PR #38, opened 2018-07-18
Fixed a problem with the ThreadLocalFstSerializer and static registration sharing between instances.

d-w-johnson:


BlueDBOnDisk/src/main/java/io/bluedb/disk/serialization/FstConfigurationFactory.java
```diff
@@ -12,6 +12,7 @@
 public class FstConfigurationFactory {
 	public static FSTConfiguration createFstConfiguration(Class<?>...registeredSerializableClasses) {
 		FSTConfiguration config = FSTConfiguration.createDefaultConfiguration();
+		config.setShareReferences(true);
```
jeremydeanlakey: we sure we want to do this? don't we want to not share references?

BlueDBOnDisk/src/main/java/io/bluedb/disk/serialization/ThreadLocalFstSerializer.java
```diff
@@ -17,33 +24,41 @@ public ThreadLocalFstSerializer(Class<?>...registeredSerializableClasses) {
 	}
 
 	@Override
-	protected FSTConfiguration initialValue() {
-		return FstConfigurationFactory.createFstConfiguration(registeredSerializableClasses);
+	protected DefaultCoder initialValue() {
+		List<Class<?>> classesToRegister = new LinkedList<>();
+		
+		classesToRegister.addAll(getClassesToAlwaysRegister());
```
jeremydeanlakey: my concern here is adding to the classes we always register.  Won't that throw off the order since the classes passed in will all get bumped one spot?

PR #39, opened 2018-07-19
Backup

jeremydeanlakey:


BlueDBOnDisk/src/main/java/io/bluedb/disk/BlueDbOnDisk.java
```diff
@@ -47,5 +72,10 @@ public void shutdown() throws BlueDbException {
 	public Path getPath() {
 		return path;
 	}
-	
+
+	protected List<BlueCollectionOnDisk<?>> getAllCollectionsFromDisk() {
+		List<File> subfolders = FileManager.getFolderContents(path.toFile(), (f) -> f.isDirectory());
+		List<BlueCollectionOnDisk<?>> collections = Blutils.map(subfolders, (File f) -> new BlueCollectionOnDisk<Serializable>(this, f.getName(), Serializable.class) );
```
cjthatcher: So, I feel like this could be done using the streams API in a cleaner way.

`List<BlueCollectionOnDisk<?>> collections = subFolders.stream().map(f -> new BlueCollectionOnDisk<Serializable>(this, f.getName()), Serializable.class)).collect(Collectors.toList());`

You could even pull the mapping function out and use it as a method reference. Maybe something like Bluetils::collectionFromFile. 

I just don't like having new code that does something that's already written.

BlueDBOnDisk/src/main/java/io/bluedb/disk/backup/BackupTask.java
```diff
@@ -0,0 +1,82 @@
...
+			copyMetaData(collection, backupPath);
+			copyDataFolders(collection, backupPath);
+		}
+		long backupEndTime = System.currentTimeMillis();
+		for (BlueCollectionOnDisk<?> collection: collectionsToBackup) {
+			copyChanges(collection, backupStartTime, backupEndTime);
+		}
+	}
+
+	private void copyMetaData(BlueCollectionOnDisk<?> collection, Path backupPath2) throws BlueDbException {
```
cjthatcher: Looks like `backupPath2` is never used in these methods. We should remove it since it is available as a final field variable.

BlueDBOnDisk/src/main/java/io/bluedb/disk/BlueDbOnDisk.java
```diff
@@ -47,5 +72,10 @@ public void shutdown() throws BlueDbException {
...
+		return collections;
```
cjthatcher: So, I feel like this could be done using the streams API in a cleaner way.

`List<BlueCollectionOnDisk<?>> collections = subFolders.stream().map(f -> new BlueCollectionOnDisk<Serializable>(this, f.getName()), Serializable.class)).collect(Collectors.toList());`

You could even pull the mapping function out and use it as a method reference. Maybe something like Bluetils::collectionFromFile. 

I just don't like having new code that does something that's already written.

BlueDBOnDisk/src/main/java/io/bluedb/disk/file/FileManager.java
```diff
@@ -135,6 +139,41 @@ public static void moveFile(Path src, BlueWriteLock<Path> lock) throws BlueDbExc
...
+			Files.copy(src, dst, StandardCopyOption.REPLACE_EXISTING);
+		} catch (IOException e) {
+			e.printStackTrace();
+			throw new BlueDbException("Can't copy '" + src + "' to '" + dst + "'.", e);
+		}
+	}
+
+	public static void copyDirectoryWithoutLock(Path src, Path dst) throws BlueDbException {
+		dst.toFile().mkdirs();
+		for (File file: src.toFile().listFiles()) {
```
cjthatcher: if src.toFile() is not a directory, line 164 will give us an NPE. I think it'd be good to check that it's a directory before we go into the loop. Safety first kids.

BlueDBOnDisk/src/main/java/io/bluedb/disk/recovery/RecoveryManager.java
```diff
@@ -8,44 +8,97 @@
...
+		filesOldEnoughToDelete.forEach((File f) -> f.delete());
+		lastLogCleanup = System.currentTimeMillis();
+	}
+
+	public List<File> getChangeHistory(long backupStartTime, long backupEndTime) throws BlueDbException {
+		List<File> files = FileManager.getFolderContents(historyFolderPath, SUFFIX);
+		if (files.isEmpty()) {
+			return files;
+		}
+		Blutils.sortByMappedvalue(files, (File f) -> extractTimestamp(f) );
```
cjthatcher: You should camel-cap the whole word. `sortByMappedValue`. 

Your `value` is lower case right now.

BlueDBOnDisk/src/main/java/io/bluedb/disk/recovery/RecoveryManager.java
```diff
@@ -71,4 +123,24 @@ public void recover() {
...
 		}
 	}
+
+	public static Long extractTimestamp(File file) {
+		try {
+			String fileName = file.getName();
+			String longString = fileName.split("[.]")[0];
+			return Long.valueOf(longString);
+		} catch (Throwable t) {
+			return null;
```
cjthatcher: returning null here can have some bad consequences. If we use the primitive without checking for `null` in the boxed value, we can get an NPE that wrecks us. 

We are occasionally checking this value with a `<` without checking it. I'd recommend returning some marker value, like -1. Or, failing that, we might go crazy and return an Optional<Long>.

BlueDBOnDisk/src/main/java/io/bluedb/disk/backup/BackupTask.java
jeremydeanlakey: removed

BlueDBOnDisk/src/main/java/io/bluedb/disk/file/FileManager.java
jeremydeanlakey: done

BlueDBOnDisk/src/main/java/io/bluedb/disk/recovery/RecoveryManager.java
jeremydeanlakey: done

BlueDBOnDisk/src/main/java/io/bluedb/disk/recovery/RecoveryManager.java
jeremydeanlakey: so I filtered out the ones with null timestamps.  if you don't like the change, maybe we should discuss

BlueDBOnDisk/src/main/java/io/bluedb/disk/BlueDbOnDisk.java
jeremydeanlakey: How strongly do you feel about that?  In general, I find the streams API to be clunky - it clutters up the code a lot.

BlueDBOnDisk/src/main/java/io/bluedb/disk/recovery/TimeStampedFile.java
```diff
@@ -0,0 +1,36 @@
...
+		return timestamp.compareTo(o.timestamp);
+	}
+
+	public static Long extractTimestamp(File file) {
+		try {
+			String fileName = file.getName();
+			String longString = fileName.split("[.]")[0];
+			return Long.valueOf(longString);
+		} catch (Throwable t) {
+			return null;
```
cjthatcher: I still don't love using "null" as a Long value. It feels risky. I'm okay to merge it, but I do think it'd be worth considering using a special case value like -1. But I don't care enough to hold up the merge.

PR #42, opened 2018-07-23
Refinements, test coverage, backup fixes

jeremydeanlakey:


BlueDBOnDisk/src/main/java/io/bluedb/disk/BlueDbOnDisk.java
```diff
@@ -81,6 +81,12 @@ public Path getPath() {
...
 	@SuppressWarnings({"rawtypes", "unchecked"})
 	protected BlueCollectionOnDisk getCollection(String folderName) throws BlueDbException {
-		return new BlueCollectionOnDisk(this, folderName, Serializable.class);
+		synchronized (collections) {
+			BlueCollectionOnDisk collection = collections.get(folderName);
+			if (collection == null) {
+				collection = new BlueCollectionOnDisk(this, folderName, Serializable.class);
+			}
+			return collection;
+		}
```
d-w-johnson: If you look at the other getCollection method this isn't doing everything it needs to. For example, its not adding the collection to the map. What is this one used for? Is it only for tests?

jeremydeanlakey: It's only for backup.  Although perhaps it should add to the map to remove the chances of a duplicate collection being created

jeremydeanlakey: Actually, if we put it in the map, it will have the type Serializable which will throw exceptions if we try to get it with another type later.

BlueDBOnDisk/src/main/java/io/bluedb/disk/recovery/RecoveryManager.java
```diff
@@ -96,20 +96,10 @@ protected void cleanupHistory() throws BlueDbException {
...
-			}
-			if (timeStampedFile.getTimestamp() <= backupEndTime) {
-				lastChangeToKeep = i;
-			}
-		}
-		
+		// Note: the last change before backup might by incomplete so we'll include it
+		int lastChangeBeforeBackup = Blutils.lastIndex(timestampedFiles, (t) -> t.getTimestamp() < backupStartTime );
+		int firstChangeToKeep = Math.max(0,  lastChangeBeforeBackup);
+		int lastChangeToKeep = Blutils.lastIndex(timestampedFiles, (t) -> t.getTimestamp() <= backupEndTime );
```
d-w-johnson: You could get the lastChangeToKeep first then pass in the index to start at for lastChangeBeforeBackup to avoid wasting checking items that you know are outside the time you're looking for. There probably isn't ever going to be enough items to really matter though right?

BlueDBOnDisk/src/main/java/io/bluedb/disk/file/FileManager.java
```diff
@@ -71,6 +71,24 @@ public void lockMoveFileUnlock(Path src, Path dst) throws BlueDbException {
...
+		} catch (Throwable t) { // make damn sure we don't hold onto the lock
+			lock.release();
+			throw new BlueDbException("Error attempting to acquire read lock", t);
+		}
+		lock.release();
+		return null;
+	}
+
+	public boolean exists(Path path) {
+		return path.toFile().exists();
```
d-w-johnson: I haven't mentioned things like this before and you don't need to change this. But there is a Files.exists(Path path) method and a Files.createDirectories method that allow you to use the path object instead of getting the file. Doesn't matter that much. Now that we use path objects more I like using these methods instead of the java.io.File ones

jeremydeanlakey: I actually created this partly to make sure I could generate an exception and test how it's handled.

BlueDBOnDisk/src/main/java/io/bluedb/disk/recovery/RecoveryManager.java
jeremydeanlakey: I don't think efficiency matters much here.  It's only going to iterate over maybe a few thousand every time you backup.

BlueDBOnDisk/src/main/java/io/bluedb/disk/Blutils.java
```diff
@@ -67,6 +67,17 @@ public int compare(X x1, X x2) {
 		return results;
 	}
 
+	public static <X> int lastIndex(List<X> values, Predicate<X> condition) {
```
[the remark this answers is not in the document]
jeremydeanlakey: The values passed in are ordered where I actually use this.  However, it was hard to communicate to any future user of this method that they must be ordered so instead of I just literally check the lastIndex matching the condition whether or not it's ordered.

PR #44, opened 2018-07-25
key types and folder structures

jeremydeanlakey:
- require key type when getting collection (and throw exception if wrong key is used)
- save key type in meta data
- break out SegmentPathManager (non time and time)
- changes to Keys and grouping numbers

also:
- better function name for getting untyped collection for backup
- remove ambiguous TimeKey constructor


BlueDBOnDisk/src/main/java/io/bluedb/disk/collection/BlueCollectionOnDisk.java
```diff
@@ -47,16 +47,16 @@
 	private final RollupScheduler rollupScheduler;
 	private final CollectionMetaData metaData;
 
-	public BlueCollectionOnDisk(BlueDbOnDisk db, String name, Class<T> valueType, Class<? extends BlueKey> keyType2) throws BlueDbException {
+	public BlueCollectionOnDisk(BlueDbOnDisk db, String name, Class<T> valueType, Class<? extends BlueKey> keyType) throws BlueDbException {
```
d-w-johnson: I would consider putting the key first. Usually key comes before value. Also, it would make sense to allow them to pass in extra classes to register after the value. Then the value and classes to register will be next to each other in the parameters list.

d-w-johnson: Also, I'm wondering if it would make more sense to have a declareCollection method and a getCollection method. The former specifies the name, keyType, valueType, and classes to register. The latter just specifies the name. I know they can create a method so they only have to get the collection once, but it seems a little weird to pass in so many parameters every time they get the collection.

BlueDBOnDisk/src/main/java/io/bluedb/disk/collection/CollectionMetaData.java
```diff
@@ -15,13 +16,15 @@
...
 	private static final String FILENAME_SERIALIZED_CLASSES = "serialized_classes";
+	private static final String FILENAME_KEY_TYPE = "key_type";
 	private static final String META_DATA_FOLDER = ".meta";
 	
 	final Path folderPath;
 	final FileManager fileManager;
 	final Path maxIntegerPath;
 	final Path maxLongPath;
 	final Path serializedClassesPath;
+	final Path keyTypePath;
```
d-w-johnson: You could generate these paths on demand right? We don't really need to have the cached all the time?

jeremydeanlakey: Yeah.  Because some of these are accessed frequently, I calculated their paths up front (also makes methods easier to read).  However, it's not really necessary to cache them.  On the other hand, it's one per collection so it doesn't take much memory.  I don't really care strongly either way.

BlueDBOnDisk/src/main/java/io/bluedb/disk/collection/BlueCollectionOnDisk.java
jeremydeanlakey: I think these are good ideas and implemented both.

BlueDBInMemory/src/main/java/io/bluedb/memory/BlueDbInMemory.java
```diff
@@ -32,9 +32,18 @@ public BlueDbInMemory(Path directory) throws BlueDbException {
 		initializeFromDirectory();
 	}
 
+	@Override
+	public <T extends Serializable> BlueCollection<T> getCollection(String name) throws BlueDbException {
+		synchronized(collections) {
+			@SuppressWarnings("unchecked")
+			BlueCollection<T> collection = (BlueCollection<T>)(collections.get(name));
```
d-w-johnson: Do we want to do this in a try and trow a BlueDbException with a caused by? Or do we want to add `throws ClassCastException` to the method signature?

BlueDBInMemory/src/main/java/io/bluedb/memory/BlueDbInMemory.java
```diff
@@ -32,9 +32,18 @@ public BlueDbInMemory(Path directory) throws BlueDbException {
...
+			return collection;
+		}
+	}
+
 	@SuppressWarnings("unchecked")
 	@Override
-	public <T extends Serializable> BlueCollection<T> getCollection(Class<? extends BlueKey> keyType, Class<T> type, String name) throws BlueDbException {
+	public <T extends Serializable> BlueCollection<T> initializeCollection(String name, Class<? extends BlueKey> keyType, Class<T> type, Class<?>... additionalClassesToRegister) throws BlueDbException {
```
d-w-johnson: Maybe name the `type` parameter `valueType` instead?

PR #46, opened 2018-08-03
separate and clarify path managers into time, long, and int

jeremydeanlakey:


BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/SegmentManager.java
```diff
@@ -78,7 +78,7 @@ protected static SegmentPathManager createSegmentPathManager(Path collectionPath
 			return new TimeSegmentPathManager(collectionPath);
 		} else if (LongKey.class.isAssignableFrom(keyType)) {
 			return new LongSegmentPathManager(collectionPath);
-		} else if (ValueKey.class.isAssignableFrom(keyType)) {
+		} else if (HashGroupedKey.class.isAssignableFrom(keyType)) {
```
d-w-johnson: I think I'd rather have the path manager be called something like HashGroupedSegmentPathManager since LongKey is timeless but is not managed by the NonTimeSegmentPathManager

PR #47, opened 2018-08-06
Cleanup history faster but not while doing backups

jeremydeanlakey:


BlueDBOnDisk/src/main/java/io/bluedb/disk/BlueDbOnDisk.java
```diff
@@ -63,12 +63,15 @@
 
 	@Override
 	public void backup(Path zipPath) throws BlueDbException {
```
d-w-johnson: Do we have any protection against them doing multiple backups at the same time? Maybe that's not worth doing right now. Would it cause problems? At the very least, the first backup would change all of the retention periods back to default while the second one was still running.

BlueDBOnDisk/src/main/java/io/bluedb/disk/BlueDbOnDisk.java
```diff
@@ -63,12 +63,15 @@
...
+		BackupTask backupTask = new BackupTask(this, zipPath);
+		List<BlueCollectionOnDisk<?>> collectionsToBackup = getAllCollectionsFromDisk();
 		try {
-			BackupTask backupTask = new BackupTask(this, zipPath);
-			List<BlueCollectionOnDisk<?>> collectionsToBackup = getAllCollectionsFromDisk();
+			long twoHours = TimeUnit.HOURS.toMillis(2);
+			collectionsToBackup.forEach((c) -> (c).getRecoveryManager().setRetentionPeriod(twoHours));
```
d-w-johnson: If history is cleaned up right before the default is changed will that cause a problem with the backup?

PR #63, opened 2018-09-07
Added support for UUIDKey. Added it to tests everywhere that StringKey

d-w-johnson:
was referenced.


jeremydeanlakey:
Looks good.

PR #82, opened 2018-10-23
Fix failing test

jeremydeanlakey:
Make test work on jenkins and in development.

There's probably a cleaner way to do this, but the way we fetch the file doesn't matter much so it's probably not worth improving.

### DIFF
--- a/BlueDBOnDisk/src/test/java/io/bluedb/disk/backup/BackupManagerTest.java
+++ b/BlueDBOnDisk/src/test/java/io/bluedb/disk/backup/BackupManagerTest.java
@@ -3,6 +3,7 @@ package io.bluedb.disk.backup;
 import java.io.File;
 import java.io.IOException;
 import java.net.URISyntaxException;
+import java.net.URL;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


BlueDBOnDisk/src/test/java/io/bluedb/disk/backup/BackupManagerTest.java
```diff
@@ -148,6 +149,15 @@ private BlueDbOnDisk createTestRestoreDatabase() throws BlueDbException {
 	}
 
 	private Path getbackedUpZipPath() throws URISyntaxException {
-		return Paths.get(this.getClass().getResource("/backup.zip").toURI());
+		URL url = this.getClass().getResource("/backup.zip");
```
d-w-johnson: There should be a better way to do this. Is this to get it to work in eclipse then? I can look at it

d-w-johnson: Or maybe I can just look at it with you tomorrow. This works on my machine both in gradle and in eclipse the way it was before these commits.

jeremydeanlakey: this is to get the test to run in eclipse.  There probably is a better way to do it but since it's a one-time-use, well-isolated helper function in a unit test, I didn't think it was worth the time to do better.  If you want to look it over together, I'm open to doing so.

d-w-johnson: Its just that I fixed this 3 weeks ago and it works in my Eclipse. And it follows the pattern that we are now using in Chronicall w/ Gradle. So I'm wondering if something just isn't set up right in your Eclipse. Plus if we can avoid having a complex way to get resources that will make future tests easier to write.

PR #86, opened 2018-11-21
Updated hash key segment path levels

d-w-johnson:


jeremydeanlakey:
I still want to get back up to 100% code coverage asap.  (I'll do it during the break even).  But I think we can probably survive with the few untested lines of code.

PR #89, opened 2018-12-11
Adds the handling of arrays to the reflective object validation method

d-w-johnson:


BlueDBOnDisk/src/test/java/io/bluedb/disk/serialization/ThreadLocalFstSerializerTest.java
```diff
@@ -57,9 +57,9 @@ public void testAddingRegisteredClass() throws SerializationException {
 	
 	@Test
 	public void testDeserializationIssue() {
-		int testCount = 15;
-		int mutationCountPerTest = 30;
-		int callCountPerTest = 1000;
+		int testCount = 200;
+		int mutationCountPerTest = 15;
+		int callCountPerTest = 100;
```
d-w-johnson: This test was taking like 45 minutes to complete. The cause of the slowdown is that every time I mutate an object I end up adding data to it so validating that object takes longer and longer throughout the test. I noticed that when I remove the fix the test usually fails pretty quick. So I figured if I ran more small tests then I'd make the test pass quicker while still ensuring that it fails without the fix.

PR #95, opened 2019-04-17
batch updates

jeremydeanlakey:


BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
```diff
@@ -175,6 +229,17 @@ private void moveRolledUpFileAndDeleteSourceFiles(Path newRolledupPath, Path tem
...
 
+	public static List<Range> getAllFileRangesInOrder(Path segmentPath) {
+		File segmentFolder = segmentPath.toFile();
+		List<File> allFilesInSegment = FileManager.getFolderContents(segmentFolder);
+		List<Range> allFileRanges = allFilesInSegment.stream()
+				.map( (File fl) -> Range.fromUnderscoreDelmimitedString(fl.getName()) )
+				.filter((r) -> r != null)
+				.collect(Collectors.toList());
+		Collections.sort(allFileRanges);
+		return allFileRanges;
```
d-w-johnson: I'd do the stream like this (in case you're interested):
```
List<Range> allFileRanges = allFilesInSegment.stream()
				.map( (File fl) -> Range.fromUnderscoreDelmimitedString(fl.getName()) )
				.filter(Objects::nonNull)
				.sorted()
				.collect(Collectors.toList());
```

d-w-johnson: Or better yet, you could make the Range.from method take a File object and could make it look like this:
```
List<Range> allFileRanges = allFilesInSegment.stream()
				.map(Range::fromFileWithUnderscoreDelmimitedName)
				.filter(Objects::nonNull)
				.sorted()
				.collect(Collectors.toList());
```

BlueDBOnDisk/src/main/java/io/bluedb/disk/BatchUtils.java
```diff
@@ -0,0 +1,67 @@
...
+			IndividualChange<T> nextChange = iterator.next();
+			if (!nextChange.overlaps(beyondSegment)) {
+				iterator.remove();
+			} else if (!nextChange.overlaps(segmentRange)) {
+				return;  // we've past all the changes in this segment
+			}
+		}
+	}
+
+	protected static <T extends Serializable> LinkedList<IndividualChange<T>> getChangesOverlappingSegment( List<IndividualChange<T>> sortedChanges, Segment<T> segment) {
```
d-w-johnson: This isn't used anymore. It also assumes that you don't have anything that is before the given segment which is not intuitive based on the name.

BlueDBOnDisk/src/main/java/io/bluedb/disk/collection/index/BlueIndexOnDisk.java
```diff
@@ -112,10 +124,37 @@ public void rollup(Range range) throws BlueDbException {
...
+		BlueKey destinationKey = change.getKey();
+		T newValue = change.getNewValue();
+		return toCompositeKeys(destinationKey, newValue);
+	}
+
+	private List<IndexCompositeKey<I>> toCompositeKeys(BlueKey destination, T newItem) {
 		List<I> indexKeys = keyExtractor.extractKeys(newItem);
 		CheckedFunction<I, IndexCompositeKey<I>> indexToComposite = (indexKey) -> new IndexCompositeKey<I>(indexKey, destination);
-		List<IndexCompositeKey<I>> compositeKeys = Blutils.map(indexKeys, indexToComposite);
+		List<IndexCompositeKey<I>> compositeKeys = Blutils.mapIgnoringExceptions(indexKeys, indexToComposite);
```
d-w-johnson: Why are you no longer throwing BlueDbException? How does that affect the places that were calling that before? (Add and Remove)

BlueDBOnDisk/src/main/java/io/bluedb/disk/collection/task/BatchChangeTask.java
```diff
@@ -0,0 +1,45 @@
...
+	}	
+
+	@Override
+	public String toString() {
+		return "<" + getClass().getSimpleName() + " for " + sortedChanges.size() + ">";
+	}
+
+	private static <T extends Serializable> List<IndividualChange<T>> toSortedChangeList(Map<BlueKey, T> values) {
+		return values.entrySet().stream()
+				.map( (e) -> IndividualChange.insert(e.getKey(), e.getValue()) )
```
d-w-johnson: `IndividualChange.insert` doesn't seem like its creating a change object. I'd prefer if the method was named something like `createInsertChange`

BlueDBOnDisk/src/main/java/io/bluedb/disk/recovery/IndividualChange.java
```diff
@@ -0,0 +1,59 @@
...
+
+	private final BlueKey key;
+	private final T oldValue;
+	private final T newValue;
+
+	public static <T extends Serializable> IndividualChange<T> insert(BlueKey key, T value) {
+		return new IndividualChange<T>(key, null, value);
+	}
+
+	public IndividualChange(BlueKey key, T oldValue, T newValue) {
```
d-w-johnson: Looks like it'll be pretty easy to make a batch update or batch delete using this class. Is that accurate? I like it!

BlueDBOnDisk/src/main/java/io/bluedb/disk/BatchUtils.java
jeremydeanlakey: this method was moved around a few times.  It's original name depended somewhat on context which is no longer there.  I'll just delete it

BlueDBOnDisk/src/main/java/io/bluedb/disk/collection/index/BlueIndexOnDisk.java
jeremydeanlakey: Blutils.map was created to handle mappers that can throw exceptions and Blutils.mapIgnoringExceptions was created to handle those and ignore the exceptions.  I used it because I needed a mapper that didn't throw exceptions and Java streams is so clunky.  But I think mapIgnoringExceptions  is not appropriate.

I replaced a couple of these calls with a helper function and another one with a stream

BlueDBOnDisk/src/main/java/io/bluedb/disk/collection/task/BatchChangeTask.java
jeremydeanlakey: sounds reasonable, changing

BlueDBOnDisk/src/main/java/io/bluedb/disk/recovery/IndividualChange.java
jeremydeanlakey: Yes, by design

BlueDBOnDisk/src/main/java/io/bluedb/disk/segment/Segment.java
jeremydeanlakey: sounds good, making changes.

PR #96, opened 2019-04-22
Bluekey compareto

jeremydeanlakey:


BlueDB/src/main/java/io/bluedb/api/keys/HashGroupedKey.java
```diff
@@ -1,11 +1,67 @@
...
+			return true;
+		}
+		if (obj == null) {
+			return false;
+		}
+		if (getClass() != obj.getClass()) {
+			return false;
+		}
+		@SuppressWarnings("unchecked")
+		HashGroupedKey<T> other = (HashGroupedKey<T>) obj;
```
d-w-johnson: Just cause something is a HashGroupedKey doesn't mean that its a HashGroupedKey<T>. What if the generic types are different in the objects? Could you cast it to HashGroupedKey<?> instead?

jeremydeanlakey: Look 3 lines above.  We explicitly check if the class is the same.

jeremydeanlakey: Well, I guess it's a little more complicated, because you could subclass it with another generic, and instantiate two different instances.  I changed this class to generic so that subclasses (which I haven't made generic) could use final methods with standardized logic to reduce risk of bugs.  I don't have intentions of making generic subclasses.  But I think your approach is strictly superior, so I'll make the change.

BlueDB/src/main/java/io/bluedb/api/keys/HashGroupedKey.java
```diff
@@ -1,11 +1,66 @@
...
+		int classComparison = compareClasses(other);
+		if (classComparison != 0) {
+			return classComparison;
+		}
+		int groupingNumberComparison = Long.compare(getGroupingNumber(), other.getGroupingNumber());
+		if (groupingNumberComparison != 0) {
+			return groupingNumberComparison;
+		}
+		@SuppressWarnings("unchecked")
+		HashGroupedKey<T> otherClassed = (HashGroupedKey<T>) other;
```
d-w-johnson: This needs the same change as the .equals that you did this morning

PR #98, opened 2019-04-26
add version to metadata

jeremydeanlakey:


BlueDB/src/main/java/org/bluedb/api/BlueDbVersion.java
```diff
@@ -0,0 +1,9 @@
+package org.bluedb.api;
+
+public enum BlueDbVersion {
+	v_1_0_0,
+	v_1_0_1,
+	v_1_1_0;
+	
+	public static BlueDbVersion CURRENT = v_1_1_0;
```
d-w-johnson: Could you pull in master and update the versions. There should be 1.0.0, 1.0.1, 1.0.2, and now 2.0.0. I don't know for sure if we'll want to forever list all possible versions in here, but for now its fine. Eventually we might just want the current version and a way to parse and compare versions. That way we can check if a version is before or after a specific version without having to have an enum with every possibility.

jeremydeanlakey: Do you want to limit to major and minor?  If we make some potentially breaking change where we'll actually need to know the existing version, it should at least change minor versions.

jeremydeanlakey: Alternative, we could do major versions and add minor versions only if really necessary.

d-w-johnson: Actually, we've already tagged 2.0.0, so when this gets officially released it'll probably be 2.0.1. Sometime soon we'll need to create a checklist of tasks that need to be done when we're ready to tag a new version.

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/BlueCollectionOnDisk.java
```diff
@@ -221,6 +227,15 @@ protected void ensureCorrectKeyTypes(Collection<BlueKey> keys) throws BlueDbExce
 		}
 	}
 
+	protected static BlueDbVersion determineVersion(CollectionMetaData metaData) throws BlueDbException {
+		BlueDbVersion version = metaData.getVersion();
```
d-w-johnson: If we need to change BlueDbVersion to a String how much work is that? I don't know that I want to serialize the enum constant and have to keep track of every version we ever create.

d-w-johnson: Do we want to set legacy collections to current? Or do we want to know that they were sometime before current?

BlueDB/src/main/java/org/bluedb/api/BlueDbVersion.java
```diff
@@ -0,0 +1,9 @@
+package org.bluedb.api;
+
+public enum BlueDbVersion {
```
d-w-johnson: We decided to create an object for holding version information that we can serialize to and from disk. It will have three ints and a string. Remember to register the class. We may want to create an interface called BlueDbVersion and create a concretion called BlueDbVersionV1 so that we are free to change what information is stored in the version in the future without breaking serialization.

d-w-johnson: Oh, we also decided that existing collections with no version will be given a version of 1.0.0

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/CollectionMetaData.java
```diff
@@ -15,11 +17,13 @@
 	
 	private static final String FILENAME_SERIALIZED_CLASSES = "serialized_classes";
 	private static final String FILENAME_KEY_TYPE = "key_type";
+	private static final String FILENAME_VERSION = "bluedb_version";
```
d-w-johnson: I was originally thinking that we should make BlueDBVersion an interface so that we can create new versions of it if we need to. However, this isn't the type of a collection, its just one object stored by a specific key. There is no reason that we can't create a new key and a new type of version and have some code to convert the old one if we ever needed to change it. Its nice to have the option even though we probably won't need to change it.

I think I'm good with this merge now except for one thing. The outer build.gradle contains a version. That version is what determines the version on the jar. It needs to always match the `CURRENT` version in our code. We should probably make it so you only have to update the gradle file. If you deploy or build the project it should set the `CURRENT` version in the code to match the version in the gradle file. Or something like that.

BlueDB/src/main/java/org/bluedb/api/BlueDbVersion.java
d-w-johnson: - The version is only written once in the meta data, so its probably not imperative to register the class
- While creating an interface and versioning the concretions gives us a little more flexibility, its not that big of deal.

PR #100, opened 2019-04-30
fix getLargestEmptyRange so batch inserts are efficient

jeremydeanlakey:


BlueDBOnDisk/src/main/java/org/bluedb/disk/segment/SegmentBatch.java
```diff
@@ -57,7 +59,7 @@
 
 	protected static Range getLargestEmptyRange(List<Range> optionsSmallToBig, Set<Range> existingChunkRanges) {
 		return optionsSmallToBig.stream()
-				.sorted(Comparator.reverseOrder())
+				.sorted(reverseOrderBySize)
```
d-w-johnson: I'm curious why you had to make this change. Were `optionsSmallToBig` coming in the wrong order? Could something else be depending on that order as well? Should we fix the sorting on the `optionsSmallToBig` list?

PR #101, opened 2019-04-30
Clean empty files

jeremydeanlakey:


jeremydeanlakey:
probably do #100 first because this branched off of that.

BlueDBOnDisk/src/main/java/org/bluedb/disk/segment/SegmentBatch.java
```diff
@@ -57,7 +59,7 @@
 
 	protected static Range getLargestEmptyRange(List<Range> optionsSmallToBig, Set<Range> existingChunkRanges) {
 		return optionsSmallToBig.stream()
-				.sorted(Comparator.reverseOrder())
+				.sorted(reverseOrderBySize)
```
d-w-johnson: You should filter before you sort. Actually, you should use the min/max operation instead of findFirst again

PR #104, opened 2019-05-02
Single pre segment chunk

jeremydeanlakey:


BlueDBOnDisk/src/main/java/org/bluedb/disk/segment/Segment.java
```diff
@@ -38,20 +38,22 @@
...
 	private final List<Long> rollupLevels;
+	private final Range preSegmentRange;
 
 	public Segment(Path segmentPath, Range segmentRange, Rollupable rollupable, FileManager fileManager, final List<Long> rollupLevels) {
 		this.segmentPath = segmentPath;
 		this.segmentRange = segmentRange;
 		this.fileManager = fileManager;
 		this.rollupLevels = rollupLevels;
 		this.rollupable = rollupable;
+		this.preSegmentRange = (segmentRange==null) ? null : new Range(0, segmentRange.getStart() - 1);
```
d-w-johnson: Is null ever passed in for the segment range? Or do you expect segment range to always be set?

jeremydeanlakey: I believe it's null during some testing but it really should always be non-null in production.

PR #106, opened 2019-05-06
limit rate of rollup scheduling

jeremydeanlakey:


BlueDBOnDisk/src/main/java/org/bluedb/disk/segment/rollup/RollupScheduler.java
```diff
@@ -82,6 +82,17 @@ public boolean isRunning() {
 		return !isStopped;
 	}
 
+	protected void scheduleLimitedReadyRollups() {
+		int rollupsToSchedule = 30 - collection.getQueuedTaskCount();
+		scheduleReadyRollups(rollupsToSchedule);
```
d-w-johnson: Only schedule rollups if rollupsToSchedule is greater than 0

BlueDBOnDisk/src/main/java/org/bluedb/disk/segment/rollup/RollupScheduler.java
```diff
@@ -82,6 +82,17 @@ public boolean isRunning() {
...
+	}
+
+	protected void scheduleReadyRollups(int maxRollupsToSchedule) {
+		for (RollupTarget target: rollupTargetsReadyForRollup(maxRollupsToSchedule)) {
+			scheduleRollup(target);
+		}
+	}
+
 	protected void scheduleReadyRollups() {
```
d-w-johnson: Is this method still called? It could be dangerous to call it if we're in a big collection that hasn't been rolling up correctly

PR #108, opened 2019-05-13
BlueCollectionBuilder

jeremydeanlakey:
It probably makes sense to look commit by commit.  One of the commits is updating all the testing references.

### DIFF
--- a/BlueDB/src/main/java/org/bluedb/api/BlueCollectionBuilder.java
+++ b/BlueDB/src/main/java/org/bluedb/api/BlueCollectionBuilder.java
@@ -6,6 +6,7 @@ import org.bluedb.api.exceptions.BlueDbException;
 
 public interface BlueCollectionBuilder<T extends Serializable> {
 	public BlueCollectionBuilder<T> withName(String name);
-	public BlueCollectionBuilder<T> registerClasses(Class<? extends Serializable>[] additionalRegisteredClasses);
+	@SuppressWarnings("unchecked")
+	public BlueCollectionBuilder<T> withRegisteredClasses(Class<? extends Serializable>... additionalRegisteredClasses);
 	public BlueCollection<T> build() throws BlueDbException;
 }
--- a/BlueDB/src/main/java/org/bluedb/api/BlueCollectionBuilder.java
+++ b/BlueDB/src/main/java/org/bluedb/api/BlueCollectionBuilder.java
@@ -1,12 +1,12 @@
 package org.bluedb.api;
 
 import java.io.Serializable;
+import java.util.Collection;
 
 import org.bluedb.api.exceptions.BlueDbException;
 
 public interface BlueCollectionBuilder<T extends Serializable> {
 	public BlueCollectionBuilder<T> withName(String name);
-	@SuppressWarnings("unchecked")
-	public BlueCollectionBuilder<T> withRegisteredClasses(Class<? extends Serializable>... additionalRegisteredClasses);
+	public BlueCollectionBuilder<T> usingClasses(Collection<Class<? extends Serializable>> classesToRegister);
 	public BlueCollection<T> build() throws BlueDbException;
 }
--- a/BlueDB/src/main/java/org/bluedb/api/BlueCollectionBuilder.java
+++ b/BlueDB/src/main/java/org/bluedb/api/BlueCollectionBuilder.java
@@ -6,7 +6,6 @@ import java.util.Collection;
 import org.bluedb.api.exceptions.BlueDbException;
 
 public interface BlueCollectionBuilder<T extends Serializable> {
-	public BlueCollectionBuilder<T> withName(String name);
-	public BlueCollectionBuilder<T> usingClasses(Collection<Class<? extends Serializable>> classesToRegister);
+	public BlueCollectionBuilder<T> withOptimizedClasses(Collection<Class<? extends Serializable>> classesToRegister);
 	public BlueCollection<T> build() throws BlueDbException;
 }
--- a/BlueDB/src/main/java/org/bluedb/api/BlueCollectionBuilder.java
+++ b/BlueDB/src/main/java/org/bluedb/api/BlueCollectionBuilder.java
@@ -1,0 +1,11 @@
+package org.bluedb.api;
+
+import java.io.Serializable;
+
+import org.bluedb.api.exceptions.BlueDbException;
+
+public interface BlueCollectionBuilder<T extends Serializable> {
+	public BlueCollectionBuilder<T> withName(String name);
+	public BlueCollectionBuilder<T> registerClasses(Class<? extends Serializable>[] additionalRegisteredClasses);
+	public BlueCollection<T> build() throws BlueDbException;
+}
--- a/BlueDB/src/main/java/org/bluedb/api/BlueDb.java
+++ b/BlueDB/src/main/java/org/bluedb/api/BlueDb.java
@@ -9,7 +9,7 @@ public interface BlueDb {
 	
 	public <T extends Serializable> BlueCollection<T> initializeCollection(String name, Class<? extends BlueKey> keyType, Class<T> valueType, @SuppressWarnings("unchecked") Class<? extends Serializable>... additionalClassesToRegister) throws BlueDbException;
 
-	public <T extends Serializable, K extends BlueKey> BlueCollectionBuilder<T> collectionBuilder(Class<K> keyType, Class<T> valueType);
+	public <T extends Serializable, K extends BlueKey> BlueCollectionBuilder<T> collectionBuilder(String name, Class<K> keyType, Class<T> valueType);
 
 	public <T extends Serializable> BlueCollection<T> getCollection(String name, Class<T> valueType) throws BlueDbException;
 
--- a/BlueDB/src/main/java/org/bluedb/api/BlueDb.java
+++ b/BlueDB/src/main/java/org/bluedb/api/BlueDb.java
@@ -9,6 +9,8 @@ public interface BlueDb {
 	
 	public <T extends Serializable> BlueCollection<T> initializeCollection(String name, Class<? extends BlueKey> keyType, Class<T> valueType, @SuppressWarnings("unchecked") Class<? extends Serializable>... additionalClassesToRegister) throws BlueDbException;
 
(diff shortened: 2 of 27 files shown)


BlueDB/src/main/java/org/bluedb/api/BlueCollectionBuilder.java
```diff
@@ -0,0 +1,11 @@
+package org.bluedb.api;
+
+import java.io.Serializable;
+
+import org.bluedb.api.exceptions.BlueDbException;
+
+public interface BlueCollectionBuilder<T extends Serializable> {
+	public BlueCollectionBuilder<T> withName(String name);
+	public BlueCollectionBuilder<T> registerClasses(Class<? extends Serializable>[] additionalRegisteredClasses);
```
d-w-johnson: I like something like `classesToRegister` better than `additionalRegisteredClasses` for the parameter name. Unless we really want them to know that they don't need to register things like standard data structures.

BlueDBOnDisk/src/main/java/org/bluedb/disk/BlueCollectionOnDiskBuilder.java
```diff
@@ -0,0 +1,42 @@
...
+import org.bluedb.api.exceptions.BlueDbException;
+import org.bluedb.api.keys.BlueKey;
+import org.bluedb.disk.collection.BlueCollectionOnDisk;
+
+public class BlueCollectionOnDiskBuilder<T extends Serializable> implements BlueCollectionBuilder<T>{
+
+	private final BlueDbOnDisk db;
+	private final Class<T> valueType;
+	private final Class<? extends BlueKey> requestedKeyType;
+	String name;
```
d-w-johnson: Name is a required parameter so I'm wondering if we should set it in the constructor instead of having the `withName` method. As it stands now, if they don't call `withName` is it going to just throw an exception? I'd rather make it more obvious that you have to give it a name.

BlueDBOnDisk/src/main/java/org/bluedb/disk/BlueCollectionOnDiskBuilder.java
```diff
@@ -28,16 +29,18 @@ protected BlueCollectionOnDiskBuilder(BlueDbOnDisk db, Class<? extends BlueKey>
 		return this;
 	}
 
-	@SuppressWarnings("unchecked")
 	@Override
-	public BlueCollectionOnDiskBuilder<T> withRegisteredClasses(Class<? extends Serializable>... additionalRegisteredClasses) {
-		this.additionalRegisteredClasses = additionalRegisteredClasses;
+	public BlueCollectionOnDiskBuilder<T> usingClasses(Collection<Class<? extends Serializable>> classesToRegister) {
```
d-w-johnson: I'm not sure how I feel about the `usingClasses` name. That doesn't really communicate why it might be important to the user to call this. `register` doesn't really either though...

BlueDB/src/main/java/org/bluedb/api/BlueCollectionBuilder.java
jeremydeanlakey: that change is made in a later request.

BlueDB/src/main/java/org/bluedb/api/BlueDb.java
```diff
@@ -9,6 +9,8 @@
 	
 	public <T extends Serializable> BlueCollection<T> initializeCollection(String name, Class<? extends BlueKey> keyType, Class<T> valueType, @SuppressWarnings("unchecked") Class<? extends Serializable>... additionalClassesToRegister) throws BlueDbException;
```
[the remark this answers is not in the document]
jeremydeanlakey: Yeah, I think we should deprecate it soon.

I prefer collectionBuilder to buildCollection because buildCollection seems to imply action taken when action isn't taken until build() is called.  I'd be open for discussion or other ideas though.

BlueDBOnDisk/src/main/java/org/bluedb/disk/BlueCollectionOnDiskBuilder.java
jeremydeanlakey: The main reason I want a builder is to take out optional things (like the varargs classes or the future segment size options).  I suppose we could require a name when you instantiate the builder to prevent run time exceptions.

Do you want collectionBuilder(...) to have all required parameters and optional parameters added in other builder methods?

BlueDBOnDisk/src/main/java/org/bluedb/disk/BlueCollectionOnDiskBuilder.java
jeremydeanlakey: I'd be happy to use a better name if you can think of one.

BlueDBOnDisk/src/main/java/org/bluedb/disk/BlueCollectionOnDiskBuilder.java
d-w-johnson: That's what I was thinking. I don't think we'll really be adding required things from now on, just optional configuration and stuff. I think I'd rather have the extra parameter than for the user to have to remember to set the name.

BlueDBOnDisk/src/main/java/org/bluedb/disk/BlueCollectionOnDiskBuilder.java
d-w-johnson: rename to `withOptimizedClasses`

BlueDBOnDisk/src/main/java/org/bluedb/disk/BlueDbOnDisk.java
```diff
@@ -107,7 +113,7 @@ protected BlueCollectionOnDisk getUntypedCollectionForBackup(String folderName)
 			collection = collections.get(folderName);
 		}
 		if (collection == null) {
-			collection = new BlueCollectionOnDisk(this, folderName, null, Serializable.class);
+			collection = new BlueCollectionOnDisk(this, folderName, null, Serializable.class, Arrays.asList());
```
d-w-johnson: You meant to pass in an empty list here right? I think you did, just want to make sure

jeremydeanlakey: yeah

BlueDBOnDisk/src/test/java/org/bluedb/disk/collection/CollectionMetaDataTest.java
```diff
@@ -74,13 +74,13 @@ public void test_getAndAddToSerializedClassList() throws Exception {
...
-		Class<? extends Serializable>[] afterAdding1 = metaData.getAndAddToSerializedClassList(TestValue.class);
+		Class<? extends Serializable>[] afterAdding1 = metaData.getAndAddToSerializedClassList(TestValue.class, Arrays.asList());
 		assertArrayEquals(testValue1PlusDefaults, afterAdding1);
 
-		Class<? extends Serializable>[] afterAdding1Again = metaData.getAndAddToSerializedClassList(TestValue.class);
+		Class<? extends Serializable>[] afterAdding1Again = metaData.getAndAddToSerializedClassList(TestValue.class, Arrays.asList());
 		assertArrayEquals(testValue1PlusDefaults, afterAdding1Again);
 
-		Class<? extends Serializable>[] afterAddingBoth = metaData.getAndAddToSerializedClassList(TestValue.class, TestValue2.class);
+		Class<? extends Serializable>[] afterAddingBoth = metaData.getAndAddToSerializedClassList(TestValue.class, Arrays.asList(TestValue2.class));
```
d-w-johnson: Do we have tests using the builder instead of the deprecated method? I'd rather update most stuff to use the builder at some point. Doesn't have to be now.

BlueDBOnDisk/src/test/java/org/bluedb/disk/BlueDbOnDiskTest.java
```diff
@@ -77,22 +77,31 @@ public void test_initializeCollection_existing_correct_type() throws Exception {
...
 		} catch(BlueDbException e) {
 		}
 	}
 
-	@SuppressWarnings("unchecked")
 	@Test
 	public void test_initializeCollection_invalid_key_type() {
 		try {
-			db.initializeCollection(getTimeCollectionName(), HashGroupedKey.class, TestValue.class);
+			db.initializeCollection(getTimeCollectionName(), HashGroupedKey.class, TestValue.class, Arrays.asList());
```
d-w-johnson: Do we have tests using the builder instead of the deprecated method? I'd rather update most stuff to use the builder at some point. Doesn't have to be now.

BlueDBOnDisk/src/test/java/org/bluedb/disk/collection/CollectionMetaDataTest.java
```diff
@@ -74,13 +74,13 @@ public void test_getAndAddToSerializedClassList() throws Exception {
 		Class<? extends Serializable>[] testValue1PlusDefaults = concatenate(defaultClasses, testValue1);
 		Class<? extends Serializable>[] testValueBothPlusDefaults =  concatenate(defaultClasses, testValueBoth);
 
-		Class<? extends Serializable>[] afterAdding1 = metaData.getAndAddToSerializedClassList(TestValue.class);
+		Class<? extends Serializable>[] afterAdding1 = metaData.getAndAddToSerializedClassList(TestValue.class, Arrays.asList());
```
d-w-johnson: Its kind of annoying that we have to pass something now even if its just an empty list. It might be worth making a method that doesn't take in a list that calls the existing method and passes in the empty list.

BlueDBOnDisk/src/test/java/org/bluedb/disk/BlueDbOnDiskTest.java
jeremydeanlakey: I think everything uses the builder except one test that is only there to get coverage.

BlueDBOnDisk/src/test/java/org/bluedb/disk/collection/CollectionMetaDataTest.java
jeremydeanlakey: I created that method and then deleted it because it was only used for tests anyway.

PR #109, opened 2019-05-13
Blue collection.replace

jeremydeanlakey:


BlueDB/src/main/java/org/bluedb/api/BlueCollection.java
```diff
@@ -23,6 +23,8 @@
 
 	public T get(BlueKey key) throws BlueDbException;
 
+	public void replace(BlueKey key, Mapper<T> updater) throws BlueDbException;
```
d-w-johnson: I'm thinking of renaming `Mapper` to `ReplacingUpdater`

d-w-johnson: Also, doesn't `BlueQuery` need a replace method as well?

jeremydeanlakey: I like BlueQuery.replace and will implement that if there's not anything high priority for me at the moment.

PR #110, opened 2019-05-17
Flexible segment sizes

jeremydeanlakey:


BlueDB/src/main/java/org/bluedb/api/BlueCollectionBuilder.java
```diff
@@ -4,8 +4,10 @@
 import java.util.Collection;
 
 import org.bluedb.api.exceptions.BlueDbException;
+import org.bluedb.api.keys.BlueKey;
 
-public interface BlueCollectionBuilder<T extends Serializable> {
-	public BlueCollectionBuilder<T> withOptimizedClasses(Collection<Class<? extends Serializable>> classesToRegister);
+public interface BlueCollectionBuilder<K extends BlueKey, T extends Serializable> {
```
d-w-johnson: Now that we have K for key I might rather see V for value instead of T

BlueDB/src/main/java/org/bluedb/api/BlueCollectionBuilder.java
```diff
@@ -4,8 +4,10 @@
 import java.util.Collection;
 
 import org.bluedb.api.exceptions.BlueDbException;
+import org.bluedb.api.keys.BlueKey;
 
-public interface BlueCollectionBuilder<T extends Serializable> {
-	public BlueCollectionBuilder<T> withOptimizedClasses(Collection<Class<? extends Serializable>> classesToRegister);
+public interface BlueCollectionBuilder<K extends BlueKey, T extends Serializable> {
+	public BlueCollectionBuilder<K, T> withOptimizedClasses(Collection<Class<? extends Serializable>> classesToRegister);
+	public BlueCollectionBuilder<K, T> withRequestedSegmentSize(SegmentSize<K> segmentSize);
```
d-w-johnson: I don't love the word requested. It feels like bluedb could just decide to ignore you. In reality, bluedb will absolutely take it if its a new collection but won't if the collection already exists and has a different segment size. When you copy files in the Java Files API it has you pass in an optional parameter that allows you to state what to do if the files already exist. For example, you can overwrite all or skip duplicates. We could possible allow them to pass in an enum that tells us what to do. For now there would only be one option which would be the default and which would just use the existing segment size. In the future if we added better support for switching between segment sizes then we could add a migrate option without breaking the API.

BlueDB/src/main/java/org/bluedb/api/SegmentSize.java
```diff
@@ -0,0 +1,27 @@
...
+import org.bluedb.api.keys.LongKey;
+import org.bluedb.api.keys.TimeKey;
+
+public class SegmentSize<T extends BlueKey> {
+
+	public static final SegmentSize<TimeKey> TIME_1_HOUR = new SegmentSize<>(3_600_000L);
+	public static final SegmentSize<TimeKey> TIME_1_DAY = new SegmentSize<>(24 * 3_600_000L);
+	public static final SegmentSize<LongKey> LONG_DEFAULT = new SegmentSize<>(64L);
+	public static final SegmentSize<IntegerKey> INT_DEFAULT = new SegmentSize<>(256L);
+	public static final SegmentSize<HashGroupedKey<?>> HASH_DEFAULT = new SegmentSize<>(524288L);
```
d-w-johnson: I don't know that we should put default in the names. I don't know how much I like the current defaults for long, int, and hash. If we end up changing the default before end of June (I think we should consider it) then what do we call the new default? I think we should chat about what options to roll out with initially. We might want to call the current ones low density or something. And make high density options that end up being the default for new collections. Obviously legacy collections need to use the configuration that was originally the default.

BlueDB/src/main/java/org/bluedb/api/SegmentSize.java
```diff
@@ -0,0 +1,27 @@
...
+import org.bluedb.api.keys.BlueKey;
+import org.bluedb.api.keys.HashGroupedKey;
+import org.bluedb.api.keys.IntegerKey;
+import org.bluedb.api.keys.LongKey;
+import org.bluedb.api.keys.TimeKey;
+
+public class SegmentSize<T extends BlueKey> {
+
+	public static final SegmentSize<TimeKey> TIME_1_HOUR = new SegmentSize<>(3_600_000L);
+	public static final SegmentSize<TimeKey> TIME_1_DAY = new SegmentSize<>(24 * 3_600_000L);
```
d-w-johnson: We should be able to offer more than just two time options. We should check with Mike to see what he wants. I could see him using 1 Week. I could also see us using 8 hours or something like that instead of going all the way to 1 day.

mthatcher: I need to better understand the costs and benefits associated with different segment sizes. Right now my understanding is limited to:

- each time you read a file has a cost
- because the system has to deserialize everything in a given file there is cost there also

I don't know what the optimal balance is. Rather than teaching each user what that balance is, perhaps there could be some sort of SegmentBuilder that takes parameters like frequency of records, minimum query duration, maximum query duration, average query duration, or whatever else may go into choosing your optimal segment size.

BlueDBOnDisk/src/main/java/org/bluedb/disk/segment/SegmentSizeSettings.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.List;
+
+import org.bluedb.api.exceptions.BlueDbException;
+import org.bluedb.api.keys.BlueKey;
+import org.bluedb.api.keys.HashGroupedKey;
+import org.bluedb.api.keys.IntegerKey;
+import org.bluedb.api.keys.LongKey;
+import org.bluedb.api.keys.TimeKey;
+
+public enum SegmentSizeSettings {
```
d-w-johnson: I think we need to decouple our public segment size options (`SegmentSize.java`) from our private options (`SegmentSizeSettings.java`). I think that we should be persisting the private `SegmentSizeSettings` configuration in the meta data of the collection instead of just a long so that we can be sure that we give the path manager the entire complete configuration that was originally used on the collection. This gives us flexibility to have more than one private configuration option for the same public option. We could change the default configuration that is used for a particular public option without breaking pre-existing collections that are still using an old configuration. We also won't have to update the public API in order to make a change.

BlueDB/src/main/java/org/bluedb/api/SegmentSize.java
d-w-johnson: Actually, I changed my mind (see my comment on the `SegmentSizeSettings` class). I don't think the public user options need to contain specific information about their size. It should just represent what the user requested. It should be mapped to a `SegmentSizeSettings` option based on current defaults which gets persisted and contains all the information that the path manager needs. So having and option called INT_DEFAULT is fine I think. We should just have the ability to change what that means under the covers. Which may not end up using the same segment size as before.

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/BlueCollectionOnDisk.java
```diff
@@ -58,17 +59,24 @@
...
 		fileManager = new FileManager(serializer);
-		this.keyType = determineKeyType(metaData, requestedKeyType);
+		segmentSize = determineSegmentSize(metaData, requestedKeyType, segmentSize, isNewCollection);
+		keyType = determineKeyType(metaData, requestedKeyType);
+		SegmentSizeSettings segmentSizeSettings = SegmentSizeSettings.getSettings(keyType, segmentSize);
 		recoveryManager = new RecoveryManager<T>(this, fileManager, serializer);
 		rollupScheduler = new RollupScheduler(this);
-		segmentManager = new SegmentManager<T>(collectionPath, fileManager, this, this.keyType);
+		segmentManager = new SegmentManager<T>(collectionPath, fileManager, this, segmentSizeSettings);
 		indexManager = new IndexManager<>(this, collectionPath);
```
d-w-johnson: Should we support being able to specify what segment size settings to use for an index. I don't know that it needs to be configurable by the user, but we may want to set it in the index's meta data and have the ability to change the default segment size settings for different types of indices.

BlueDB/src/main/java/org/bluedb/api/BlueCollectionBuilder.java
jeremydeanlakey: sounds like a good idea

BlueDB/src/main/java/org/bluedb/api/BlueCollectionBuilder.java
jeremydeanlakey: Are you suggesting any change right now or just talking about possibly for later?

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/BlueCollectionOnDisk.java
jeremydeanlakey: probably, but I was planning to do that on a later pull request

BlueDBOnDisk/src/main/java/org/bluedb/disk/segment/SegmentSizeSettings.java
jeremydeanlakey: This is tricky.  I think we should actually commit to disk the size of the segment because we're actually already committing to that size in the folder structure.  Perhaps we shouldn't expose it in the public API though... we should discuss some time

BlueDB/src/main/java/org/bluedb/api/SegmentSize.java
d-w-johnson: Interesting idea. I think for power users who really understand BlueDB details I'll want to give the client the option to select specific sizes. But it might be nice to also support some sort of builder that auto selects based on some information you give it. I'm happy to chat about the pros and cons of different segment sizes with you. We want to write up an article that helps explain this for those who want to know more.

BlueDB/src/main/java/org/bluedb/api/BlueCollectionBuilder.java
jeremydeanlakey: given all the likely merge conflicts as well as concerns about things like documentation not matching code after merges, I'd rather wait until most of the pending pull requests are done before changing the generic parameters

BlueDB/src/main/java/org/bluedb/api/SegmentSize.java
```diff
@@ -7,21 +7,42 @@
...
 public class SegmentSize<K extends BlueKey> {
-
-	public static final SegmentSize<TimeKey> TIME_1_HOUR = new SegmentSize<>(3_600_000L);
-	public static final SegmentSize<TimeKey> TIME_1_DAY = new SegmentSize<>(24 * 3_600_000L);
-	public static final SegmentSize<LongKey> LONG_DEFAULT = new SegmentSize<>(64L);
-	public static final SegmentSize<IntegerKey> INT_DEFAULT = new SegmentSize<>(256L);
-	public static final SegmentSize<HashGroupedKey<?>> HASH_DEFAULT = new SegmentSize<>(524288L);
-
-	private final long size;
+	public static final SegmentSize<TimeKey> TIME_1_HOUR = new SegmentSize<>("TIME_1_HOUR");
```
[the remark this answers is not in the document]
d-w-johnson: Yes. intellisense will only bring up the options that match the type of the collection you're building. This is how you set it up originally.

BlueDBOnDisk/src/main/java/org/bluedb/disk/segment/SegmentSizeSetting.java
```diff
@@ -0,0 +1,169 @@
...
+	TIME_6_HOURS(new SegmentSizeConfiguration(TimeKey.class, 	asList(SIX_HOURS, 	 4L, 30L, 12L), asList(ONE_MILLI, SIX_SECONDS, ONE_HOUR, SIX_HOURS))),
+	TIME_12_HOURS(new SegmentSizeConfiguration(TimeKey.class, 	asList(TWELVE_HOURS, 2L, 30L, 12L), asList(ONE_MILLI, SIX_SECONDS, ONE_HOUR, TWELVE_HOURS))),
+	TIME_1_DAY(new SegmentSizeConfiguration(TimeKey.class, 		asList(ONE_DAY, 	     30L, 12L), asList(ONE_MILLI, SIX_SECONDS, ONE_HOUR, ONE_DAY))),
+	TIME_5_DAYS(new SegmentSizeConfiguration(TimeKey.class, 	asList(FIVE_DAYS, 		  6L, 12L), asList(ONE_MILLI, SIX_SECONDS, ONE_HOUR, ONE_DAY, FIVE_DAYS))),
+	TIME_15_DAYS(new SegmentSizeConfiguration(TimeKey.class,	asList(FIFTEEN_DAYS, 	  2L, 12L), asList(ONE_MILLI, SIX_SECONDS, ONE_HOUR, ONE_DAY, FIFTEEN_DAYS))),
+	TIME_1_MONTH(new SegmentSizeConfiguration(TimeKey.class, 	asList(ONE_MONTH, 			  12L), asList(ONE_MILLI, SIX_SECONDS, ONE_HOUR, ONE_DAY, ONE_MONTH))),
+	TIME_3_MONTHS(new SegmentSizeConfiguration(TimeKey.class,	asList(THREE_MONTHS, 		  12L), asList(ONE_MILLI, SIX_SECONDS, ONE_HOUR, ONE_DAY, ONE_MONTH, THREE_MONTHS))),
```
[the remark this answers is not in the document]
d-w-johnson: Wow, well spotted. Thanks!

BlueDBOnDisk/src/main/java/org/bluedb/disk/segment/path/SegmentPathTimeUnits.java
```diff
@@ -0,0 +1,24 @@
...
+public class SegmentPathTimeUnits {
+	public static long ONE_MILLI = 1L;
+	public static long SIX_SECONDS = 6_000L;
+	public static long ONE_MINUTE = 60_000L;
+	
+	public static long ONE_HOUR = 3_600_000L;
+	public static long TWO_HOURS = ONE_HOUR * 2;
+	public static long FOUR_HOURS = ONE_HOUR * 4;
+	public static long SIX_HOURS = ONE_HOUR * 6;
+	public static long EIGHT_HOURS = ONE_HOUR * 8;
```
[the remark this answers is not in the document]
d-w-johnson: I originally had that as a supported time segment size but removed it so that all the segment sizes could be multiples of each other. Do you want me to delete the unused ones?

BlueDBOnDisk/src/main/java/org/bluedb/disk/segment/path/SegmentSizeConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+	private final long segmentSize;
+	
+	public SegmentSizeConfiguration(Class<? extends BlueKey> keyType, List<Long> bottomToTopFolderWidths, List<Long> rollups) {
+		this.keyType = keyType;
+		this.rollupsBottomToTop = Collections.unmodifiableList(rollups);
+		this.folderSizesTopToBottom = Collections.unmodifiableList(bottomUpFolderWidthsToTopDownFolderSizes(bottomToTopFolderWidths));
+		this.segmentSize = folderSizesTopToBottom.get(folderSizesTopToBottom.size() - 1);
+	}
+
+	public static List<Long> bottomUpFolderWidthsToTopDownFolderSizes(List<Long> bottomToTopWidths) {
```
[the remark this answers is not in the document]
d-w-johnson: This is actually your code haha. I just copied it from the setting enum class into this config class when I extracted the config details into their own class. I'm cool to just leave it if you are.

BlueDBOnDisk/src/main/java/org/bluedb/disk/segment/SegmentSizeSetting.java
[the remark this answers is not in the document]
d-w-johnson: I added one more commit to change the last numbers to 4 and 2

BlueDBOnDisk/src/main/java/org/bluedb/disk/segment/path/SegmentPathTimeUnits.java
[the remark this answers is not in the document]
jeremydeanlakey: I'd prefer the unused ones be deleted, but I don't feel that strongly about it.

jeremydeanlakey:
This is looking good but looks like we have some merge conflicts now

BlueDBOnDisk/src/main/java/org/bluedb/disk/segment/path/SegmentPathTimeUnits.java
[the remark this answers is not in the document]
d-w-johnson: I deleted the unused ones

PR #111, opened 2019-05-21
Batch delete and using batches in query updates and deletes

jeremydeanlakey:


BlueDBOnDisk/src/main/java/org/bluedb/disk/recovery/IndividualChange.java
```diff
@@ -18,6 +18,10 @@
 		return new IndividualChange<T>(key, null, value);
 	}
 
+	public static <T extends Serializable> IndividualChange<T> createDeleteChange(BlueKey key) {
+		return new IndividualChange<T>(key, null, null);
```
d-w-johnson: The delete multiple has the current value as the first option and null as the second. I'm guessing this works fine for the batch delete, I just want to make sure

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/task/UpdateMultipleTask.java
```diff
@@ -28,33 +31,41 @@ public UpdateMultipleTask(BlueCollectionOnDisk<T> collection, BlueQueryOnDisk<T>
...
 		try {
-			updates = createUpdates(entities, updater);
+			changes = createChanges(entities, updater);
 		} catch(Throwable t) {
 			t.printStackTrace();
 			throw new BlueDbException("Error updating values", t);
 		}
 
+		Collections.sort(changes);
+		PendingBatchChange<T> change = PendingBatchChange.createBatchChange(changes);
```
d-w-johnson: We'll probably need to make these same changes to the replace multiple task once we merge that in

BlueDBOnDisk/src/main/java/org/bluedb/disk/recovery/IndividualChange.java
jeremydeanlakey: It should, and should be tested

PR #112, opened 2019-05-21
shared timer

jeremydeanlakey:


BlueDBOnDisk/src/main/java/org/bluedb/disk/recovery/ChangeHistoryCleaner.java
```diff
@@ -4,25 +4,30 @@
...
 import java.util.concurrent.atomic.AtomicInteger;
 import org.bluedb.disk.Blutils;
 
-public class ChangeHistoryCleaner implements Runnable {
+public class ChangeHistoryCleaner {
 
 	private static int DEFAULT_RETENTION_LIMIT = 200;
 	private int completedChangeLimit = DEFAULT_RETENTION_LIMIT;
 	private final AtomicInteger holdsOnHistoryCleanup = new AtomicInteger(0);
 	private long waitBetweenCleanups = 500;
```
d-w-johnson: Do we need to be running every 500 millis? If you have a ton of collections it might keep the timer thread pretty busy just checking if there is anything to delete all the time for all of them. The checks should be very fast if there isn't anything to delete, but I feel like we could probably raise this without causing issues.

BlueDBOnDisk/src/main/java/org/bluedb/disk/BlueDbOnDisk.java
```diff
@@ -21,6 +22,7 @@
 
 	private final Path path;
 	private final BackupManager backupManager;
+	private final Timer sharedTimer = new Timer();
```
d-w-johnson: Do you know what happens if one of your tasks throws an uncaught exception/error? Will it kill the timer thread? Will it stop that task from running again? You might want to put try catches on all your tasks. ScheduledThreadPoolExecutors are newer and the java doc is more specific with what happens in those cases.

BlueDBOnDisk/src/main/java/org/bluedb/disk/recovery/ChangeHistoryCleaner.java
jeremydeanlakey: That could be bad for busy collections running on windows because listing files of the directories is insanely slow.  But I don't think it really hurts us too much most of the time to bump it up to 5 seconds.  I'm okay doing that for now.

Some other approaches that I like better are keeping the file names so we don't have to list directory contents and/or doing the cleanup every n changes rather than every t milliseconds.  We could do those later if this becomes a problem.

BlueDBOnDisk/src/main/java/org/bluedb/disk/BlueDbOnDisk.java
jeremydeanlakey: sounds reasonable to change to ScheduledThreadPoolExecutor

PR #114, opened 2019-05-22
delete empty folders on rollup

jeremydeanlakey:


BlueDBOnDisk/src/main/java/org/bluedb/disk/segment/Segment.java
```diff
@@ -166,6 +166,9 @@ public void modifyChunk(Path targetPath, StreamingWriter<T> processor) throws Bl
 
 	public void rollup(Range timeRange) throws BlueDbException {
 		rollup(timeRange, true);
+		if (getAllFileRangesInOrder(segmentPath).size() == 0) {
```
d-w-johnson: If there are only duplicate files in the segment then the directory will not be deleted right?

jeremydeanlakey: It shouldn't get deleted if there are any files with valid names in that folder.  Which I believe is the correct behavior.  We need the duplicates in there for queries even if there's nothing that is non-duplicate.

PR #115, opened 2019-06-05
This gets rid of the one-thread-per-collection issue

mthatcher:
It replaces it with a cached thread pool that makes sure tasks are all
executed in order.

I removed an executor.shutdown() call
at io.bluedb.disk.collection.BlueCollectionOnDisk.shutdown(BlueCollectionOnDisk.java:186).
Because cached thread pools clean themselves up this will not be an
issue unless you were relying on that shutdown to prevent future task
submissions, in which case you'll need to make CachedSingleThreadingPool
handle shutdowns.

The included test verifies that it is working as expected, but may
benefit from more scrutiny.

### DIFF
--- a/BlueDBOnDisk/src/main/java/io/bluedb/disk/collection/BlueCollectionOnDisk.java
+++ b/BlueDBOnDisk/src/main/java/io/bluedb/disk/collection/BlueCollectionOnDisk.java
@@ -8,8 +8,6 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 
 import io.bluedb.api.BlueCollection;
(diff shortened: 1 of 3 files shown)


BlueDBOnDisk/src/main/java/io/bluedb/disk/collection/BlueCollectionOnDisk.java
```diff
@@ -184,7 +183,6 @@ public void shutdown() {
 		recoveryManager.getChangeHistoryCleaner().stop();
 		rollupScheduler.forceScheduleRollups();
 		rollupScheduler.stop();
-		executor.shutdown();
```
d-w-johnson: Hmm... its probably good practice to add a shutdown method on CachedWingleThreadingPool that calls shutdown on the cachedThreadPool field. That way we can be sure that we've communicated to the pool that we're shutting down. Wouldn't want threads waiting to timeout to delay the shutdown of the process.

BlueDBOnDisk/src/main/java/io/bluedb/util/CachedSingleThreadingPool.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.ThreadPoolExecutor;
+
+public class CachedSingleThreadingPool {
+
+	
+	private ConcurrentHashMap<String, ConcurrentLinkedQueue<FutureTask<Void>>> collectionTaskMap = new ConcurrentHashMap<>();
```
d-w-johnson: Looks like you're only accessing this from within synchronized blocks so we probably don't need to use concurrent classes. I might look a little more into the synchronized stuff and see if we can slim down the critical sections at all. That being said, everything inside of the synchronized blocks is super fast, so its probably fine.

d-w-johnson:
Cool, thanks Mike! This is a little behind master and we need to hash through all of the open pull requests and get them in. You may have to adjust your PR (or we can do it as well) once we're done with all of the PR's. I think everything here would work with our changes. I think we just need a way to check how many queued tasks there are for a given collection for one of our PR's. Which is easy enough to add.

d-w-johnson:
This code was based off of an old version. I got your commit into our latest code, refactored a bit, and added test coverage. It will be merged in as part of https://github.com/BlueDB/BlueDB/pull/117

PR #130, opened 2019-10-28
Fix temp file bug

jeremydeanlakey:


BlueDBOnDisk/src/main/java/org/bluedb/disk/file/FileUtils.java
```diff
@@ -21,25 +22,26 @@
...
 
-	public static List<File> getFolderContents(Path path, String suffix) {
+	public static List<File> getFolderContentsExcludingTempFiles(Path path, String suffix) {
 		FileFilter endsWithSuffix = (f) -> f.toPath().toString().endsWith(suffix);
-		return getFolderContents(path.toFile(), endsWithSuffix);
+		return getFolderContentsExcludingTempFiles(path.toFile(), endsWithSuffix);
+	}
+
+	public static List<File> getFolderContentsExcludingTempFiles(File folder, FileFilter filter) {
+		File[] folderContentsArray = folder.listFiles(filter);
```
d-w-johnson: You could pass in a filter here that checks the given filter and your own temp file filter. Just curious if there is a reason you're not doing it that way. Seems like that saves a bit of work and this is something that bluedb has to do a ton so it might actually be worth the optimization.

BlueDBOnDisk/src/main/java/org/bluedb/disk/file/FileUtils.java
```diff
@@ -21,25 +22,26 @@
...
 	}
 
-	public static List<File> getFolderContents(File folder, FileFilter filter) {
-		File[] folderContentsArray = folder.listFiles(filter);
-		if (folderContentsArray == null) {
-			return new ArrayList<>();
-		}
-		return Arrays.asList(folderContentsArray);
+	public static List<File> getFolderContentsExcludingTempFiles(File folder) {
+		File[] folderContentsArray = folder.listFiles();
```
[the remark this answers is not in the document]
jeremydeanlakey: ok

BlueDBOnDisk/src/main/java/org/bluedb/disk/file/FileUtils.java
jeremydeanlakey: ok

PR #131, opened 2019-12-05
Startup and NPE Fixes

d-w-johnson:


BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/CollectionMetaData.java
```diff
@@ -78,20 +78,32 @@ public Path getPath() {
 	public final Class<? extends Serializable>[] getAndAddToSerializedClassList(Class<? extends Serializable> primaryClass, List<Class<? extends Serializable>> additionalClasses) throws BlueDbException {
 		List<Class<? extends Serializable>> existingClassList = getSerializedClassList();
 		List<Class<? extends Serializable>> newClassList = (existingClassList == null) ? new ArrayList<>() : new ArrayList<>(existingClassList);
+		
+		boolean madeAChange = false;
```
jeremydeanlakey: couldn't you just compare the lists (existingClassList vs newClassList) at the end?

d-w-johnson: ya that would work too

PR #133, opened 2020-01-06
Versioned serialized_classes files

d-w-johnson:
This makes the serialized classes collection meta data file be versioned/timestamped 
instead of being overwritten. Knowing what order classes were registered to the collection 
is super important and this gives us a trail so that recovery is easier if something happens 
to the most recent file.

### DIFF
--- a/BlueDBOnDisk/src/test/java/org/bluedb/disk/file/FileManagerTest.java
+++ b/BlueDBOnDisk/src/test/java/org/bluedb/disk/file/FileManagerTest.java
@@ -1,5 +1,7 @@
 package org.bluedb.disk.file;
 
+import static org.junit.Assert.assertNotEquals;
+
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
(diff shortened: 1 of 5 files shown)


BlueDBOnDisk/src/test/java/org/bluedb/disk/file/FileManagerTest.java
```diff
@@ -88,7 +94,82 @@ public void test_saveObject() throws Exception {
...
+		
+		Random r = new Random();
+		SimpleDateFormat sdf = new SimpleDateFormat(FileManager.TIMESTAMP_VERSION_FORMAT);
+		
+		assertNull(FileManager.getNewestVersionPath(tempDirPath, filename));
+		
+		List<Path> testPaths = new LinkedList<Path>();
+		
+		Path legacyPath = tempDirPath.resolve(filename);
+		Files.createFile(legacyPath);
```
jeremydeanlakey: Before you add  the timestamped files, do you want to check that "assertEquals(legacyPath, FileManager.getNewestVersionPath(tempDirPath, filename));" ?

d-w-johnson: I am covering that case in the `test_saveAndLoadVersionedObjects` method, but I think you're right that it makes sense here as well. I just pushed a change for this.

PR #134, opened 2020-01-15
Rename readonly to readable + unit tests

jeremydeanlakey:


BlueDBOnDisk/src/main/java/org/bluedb/disk/ReadOnlyBlueDbOnDisk.java
```diff
@@ -40,7 +43,13 @@
 		synchronized(collections) {
 			ReadOnlyBlueCollectionOnDisk<?> untypedCollection = collections.get(name);
 			if (untypedCollection == null) {
-				return null;
+				ReadOnlyBlueCollectionOnDisk<T> collection = new ReadOnlyBlueCollectionOnDisk<>(this, name, null, valueType, Arrays.asList());
+				if (TimeKey.class.isAssignableFrom(collection.getKeyType())) {
+					Class<? extends BlueKey> keyType = collection.getKeyType();
+					collection = new ReadOnlyBlueTimeCollectionOnDisk<>(this, name, keyType, valueType, Arrays.asList());
```
d-w-johnson: Why are you creating the collection again here? What is different the second time? Is the first one shut down? Or is there a lingering executor that will stay running forever?

d-w-johnson: Oh, and looks like it doesn't try to look up the collection from the map first.

d-w-johnson: We don't want to recreate the collection every time do we? And if the collection doesn't actually exist what happens? Shouldn't we return an empty collection or something?

PR #135, opened 2020-01-21
prevent read-only from writing + handle uninitialized collections

jeremydeanlakey:


BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/BlueCollectionOnDisk.java
```diff
@@ -150,4 +151,25 @@ public void reportWrites(List<RollupTarget> rollupTargets) {
...
+
+	@Override
+	public CollectionMetaData getMetaData() {
+		return metadata;
+	}
+
+	private CollectionMetaData metadata;
+
+	@Override
+	protected CollectionMetaData getOrCreateMetadata() {
```
d-w-johnson: Why getOrCreate instead of just setting the meta data object in the constructor and returning it in the getter?

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/FacadeCollection.java
```diff
@@ -38,8 +39,19 @@ public FacadeCollection(ReadableBlueDbOnDisk db, String name, Class<T> valueType
 	}
 
 	@Override
-	public <K extends ValueKey> BlueIndex<K, T> getIndex(String name, Class<K> keyType) throws BlueDbException {
-		return getCollection().getIndex(name, keyType);
+	public <K extends ValueKey> BlueIndex<K, T> getIndex(String indexName, Class<K> indexKeyType) throws BlueDbException {
+		return new FacadeBlueIndexOnDisk<K, T>(() -> {
+			ReadableBlueCollection<T> collection = getCollection();
+			if (collection instanceof DummyReadOnlyBlueCollectionOnDisk) {
+				return null;
```
d-w-johnson: Isn't it supposed to return a dummy index here?

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/index/IndexManager.java
```diff
@@ -8,23 +8,22 @@
...
 import org.bluedb.api.keys.ValueKey;
-import org.bluedb.disk.collection.ReadOnlyBlueCollectionOnDisk;
+import org.bluedb.disk.collection.ReadWriteBlueCollectionOnDisk;
 import org.bluedb.disk.file.FileUtils;
 import org.bluedb.disk.recovery.IndividualChange;
 
-public class IndexManager<T extends Serializable> {
-
-	private static final String INDEXES_SUBFOLDER = ".index";
+public class IndexManager<T extends Serializable> extends ReadableIndexManager<T> {
```
d-w-johnson: You should probably rename this class to ReadWriteIndexManager right?

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/ReadWriteBlueCollectionOnDisk.java
```diff
@@ -0,0 +1,212 @@
...
+	public ReadWriteBlueCollectionOnDisk(ReadWriteBlueDbOnDisk db, String name, Class<? extends BlueKey> requestedKeyType, Class<T> valueType, List<Class<? extends Serializable>> additionalRegisteredClasses, SegmentSizeSetting segmentSize) throws BlueDbException {
+		super(db, name, requestedKeyType, valueType, additionalRegisteredClasses, segmentSize);
+		sharedExecutor = db.getSharedExecutor();
+		collectionKey = getPath().toString();
+		rollupScheduler = new RollupScheduler(this);
+		rollupScheduler.start();
+		fileManager = new ReadWriteFileManager(serializer);
+		recoveryManager = new RecoveryManager<T>(this, getFileManager(), getSerializer());
+		Rollupable rollupable = null;
+		rollupable = (Rollupable) this;
```
d-w-johnson: This is a bit weird how we declare the rollupable variable, set it to null, then set it to this, then pass it into the segment manager constructor. I think `Rollupable rollupable = this` should be sufficient if you don't want to pass `this` straight into the constructor. Which makes sense since passing `this` into the constructor hides the fact that you're really passing in a rollupable.

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/index/ReadWriteBlueIndexOnDisk.java
```diff
@@ -60,43 +57,60 @@
...
-		this.keyExtractor = keyExtractor;
-		this.fileManager = collection.getFileManager();
+	private ReadWriteBlueIndexOnDisk(ReadWriteBlueCollectionOnDisk<T> collection, Path indexPath, KeyExtractor<I, T> keyExtractor) throws BlueDbException {
+		super(collection, indexPath, keyExtractor);
 		this.indexName = indexPath.toFile().getName();
+		this.fileManager = collection.getFileManager();
 		SegmentSizeSetting sizeSetting = determineSegmentSize(keyExtractor.getType());
-		segmentManager = new SegmentManager<BlueKey>(indexPath, fileManager, this, sizeSetting.getConfig());
+		segmentManager = new ReadWriteSegmentManager<BlueKey>(indexPath, fileManager, this, sizeSetting.getConfig());
+		rollupScheduler = ((ReadWriteBlueCollectionOnDisk<T>) collection).getRollupScheduler();
```
d-w-johnson: This cast is unnecessary now

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/index/ReadWriteBlueIndexOnDisk.java
```diff
@@ -60,43 +57,60 @@
...
-		 * for inspiration.
-		 */
-		return SegmentSizeSetting.getOriginalDefaultSettingsFor(keyType);
+	public ReadWriteSegmentManager<BlueKey> getSegmentManager() {
+		return segmentManager;
+	}
+
+	@Override
+	public void reportReads(List<RollupTarget> rollupTargets) {
+		if (rollupScheduler != null) {
```
d-w-johnson: This null check is unnecessary. As far as I can tell, the rollup scheduler should always be set. Same goes for the null check on line 92

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/BlueCollectionOnDisk.java
jeremydeanlakey: so it needs to be created in the superclass by a method in the subclasses so that the superclass constructor can use it.  And it needs to be held in the subclass.  So we could do something like setMetadata(createMetadata()) but that seems even more awkward to me.  If you have a cleaner idea,  I'd be all for it.

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/FacadeCollection.java
jeremydeanlakey: The Supplier gets one if it can, if not, it returns null and the FacadeCollection creates a Dummy.  Alternatively, we could create a dummy here.  I see pros and cons either way.

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/index/IndexManager.java
jeremydeanlakey: did so in the latest commit

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/ReadWriteBlueCollectionOnDisk.java
jeremydeanlakey: That's pretty funny.  I'm changing it to "Rollupable rollupable = this;" in the next commit.

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/index/ReadWriteBlueIndexOnDisk.java
jeremydeanlakey: removed in the next commit

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/index/ReadWriteBlueIndexOnDisk.java
jeremydeanlakey: removing these in the next commit

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/FacadeCollection.java
d-w-johnson: This method is something that a user would call. Returning null will give them null right? Unless I'm missing something. Shouldn't we just return dummyCollection.getIndex() instead of null?

d-w-johnson: Oh sorry, we're returning the facade index here. And the facade index returns a dummy if this returns null. Got it

PR #142, opened 2020-05-14
Fixes an issue with creating read only collections that were created before we started saving segment size meta data.

d-w-johnson:
Fixes an issue with creating read only collections that were created
before we started saving segment size meta data. If those same
collections are initialized by a read/write instance of bluedb then it
will now create the meta data file.

### DIFF
--- a/BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/ReadableCollectionOnDisk.java
+++ b/BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/ReadableCollectionOnDisk.java
@@ -36,7 +36,7 @@ public abstract class ReadableCollectionOnDisk<T extends Serializable> implement
 	protected final SegmentSizeSetting segmentSizeSettings;
 
 	protected abstract ReadableCollectionMetadata getOrCreateMetadata();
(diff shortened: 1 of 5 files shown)


BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/ReadableCollectionOnDisk.java
```diff
@@ -118,17 +118,19 @@ protected void ensureCorrectKeyTypes(Collection<BlueKey> keys) throws BlueDbExce
 	}
 
 	protected static SegmentSizeSetting determineSegmentSize(ReadableCollectionMetadata metaData, Class<? extends BlueKey> keyType, SegmentSizeSetting requestedSegmentSize, boolean isNewCollection) throws BlueDbException {
-		SegmentSizeSetting existingSegmentSize = metaData.getSegmentSize();
-		if (existingSegmentSize == null) {
+		SegmentSizeSetting segmentSize = metaData.getSegmentSize();
+		if (segmentSize == null) {
 			if (!isNewCollection) {
-				return SegmentSizeSetting.getOriginalDefaultSettingsFor(keyType);
```
d-w-johnson, line 124: Was there a reason that you weren't creating the meta data file when we initialize an existing collection that doesn't already have one?

PR #147, opened 2020-07-14
Adds BlueDB support for backing up data from a specific timeframe

d-w-johnson:


BlueDBOnDisk/src/main/java/org/bluedb/disk/backup/BackupManager.java
```diff
@@ -58,39 +82,124 @@ private void copyMetaData(ReadWriteCollectionOnDisk<?> collection, Path tempFold
...
+				copyDataFileAfterFilteringBasedOnTime(segment, tempFolder, includedTimeRange, groupingNumber);
+			} else {
+				copyDataFileStraightOver(segment, tempFolder, groupingNumber);
 			}
 		}
 	}
 
-	private void copyChanges(ReadWriteCollectionOnDisk<?> collection, long backupStartTime, long backupEndTime, Path tempFolder) throws BlueDbException {
+	private void copyDataFileAfterFilteringBasedOnTime(ReadWriteSegment<?> segment, Path tempFolder, Range includedTimeRange, long groupingNumber) throws BlueDbException {
+		Path src = segment.getPathFor(groupingNumber);
```
TheSirwaffle, line 122: So in the method after this one "copyDataFileStriaghtOver" we create a lock on the path. We specifically don't do this here. What was the thinking there? Or rather when do and when do we not use locks when reading/deleting files?

BlueDBOnDisk/src/main/java/org/bluedb/disk/backup/BackupManager.java
```diff
@@ -58,39 +82,124 @@ private void copyMetaData(ReadWriteCollectionOnDisk<?> collection, Path tempFold
...
 		RecoveryManager<?> recoveryManager = collection.getRecoveryManager();
-		List<File> changesToCopy = recoveryManager.getChangeHistory(backupStartTime, backupEndTime);
+		List<File> changesToCopy = recoveryManager.getChangeHistory(includedChangeTimes.getStart(), includedChangeTimes.getEnd());
 		Path historyFolderPath = recoveryManager.getHistoryFolder();
 		Path destinationFolderPath = translatePath(dbPath, tempFolder, historyFolderPath);
 		destinationFolderPath.toFile().mkdirs();
 		for (File file: changesToCopy) {
 			Path destinationPath = Paths.get(destinationFolderPath.toString(), file.getName());
-			FileUtils.copyFileWithoutLock(file.toPath(), destinationPath);
+			if(!includedTimeRange.isMaxRange() && collection.isTimeBased()) {
```
TheSirwaffle, line 160: This is the second time I've seen this exact combination of conditionals in an if statement. It might be better to move them to a well named method.

BlueDBOnDisk/src/main/java/org/bluedb/disk/backup/BackupManager.java
```diff
@@ -58,39 +82,124 @@ private void copyMetaData(ReadWriteCollectionOnDisk<?> collection, Path tempFold
...
+				return true;
+			} else {
+				return false;
+			}
+		}
+		
+		if(change instanceof PendingBatchChange) {
+			PendingBatchChange<?> batchChange = (PendingBatchChange<?>) change;
+			batchChange.removeChangesOutsideRange(includedTimeRange);
+			if(!batchChange.isEmpty()) {
```
jeremydeanlakey, line 188: `return !batchChange.isEmpty()! `  ?

BlueDBOnDisk/src/main/java/org/bluedb/disk/backup/BackupManager.java
```diff
@@ -58,39 +82,124 @@ private void copyMetaData(ReadWriteCollectionOnDisk<?> collection, Path tempFold
...
+		Recoverable<?> change = (Recoverable<?>) collection.getFileManager().loadObject(file);
+		if(shouldCopyChange(includedTimeRange, change)) {
+			collection.getFileManager().saveObject(destinationPath, change);
 			recoveryManager.markChangePending(destinationPath);
 		}
 	}
 
+	private boolean shouldCopyChange(Range includedTimeRange, Recoverable<?> change) {
+		if(change instanceof PendingChange) {
+			if(((PendingChange<?>)change).getKey().isInRange(includedTimeRange.getStart(), includedTimeRange.getEnd())) {
```
jeremydeanlakey, line 178: Might consider making a helper method BlueKey.isInRange(TimeRange) because you do this at least twice.

BlueDBOnDisk/src/main/java/org/bluedb/disk/backup/BackupManager.java
d-w-johnson: In general, we acquire a read lock for all file reads and a write lock for all file writes. The backup process is a bit of an exception to this rule. We need a lock on the files we're reading from because other read/write queries could be attempting to read or write to/from them during the backup process. However, the files we're writing to are only touched by this process. It creates a temp directory which it writes all of the backup files to before zipping it all up. There is no need to expend resources protecting those backup files from other threads accessing them. Though the resource expenditure is probably not very significant.

Notes:
Line 127 - We open an output stream without a lock for writing to the backup file
Line 128 - We open up an input stream for the file we will read from. This method acquires a read lock for us.
Line 145 - Instead of opening an input stream that allows us to read one object at a time, we just straight up acquire a read lock so that we can copy the bytes straight over.
Line 148 - This writes the bytes to the backup file without a lock

BlueDBOnDisk/src/main/java/org/bluedb/disk/backup/BackupManager.java
d-w-johnson: Good call, I'll do that

BlueDBOnDisk/src/main/java/org/bluedb/disk/backup/BackupManager.java
d-w-johnson: 👍

BlueDBOnDisk/src/main/java/org/bluedb/disk/backup/BackupManager.java
d-w-johnson: Hey Jeremy what's up? Always glad to see your comments! I just tried to implement this comment but BlueKey is an interface in the API project and it doesn't know about the Range class so I don't think I'm going to worry too much about it for now.

jeremydeanlakey: :) How goes it?   Yeah, I was waiting for a build so I figured I'd scroll through and take a brief look.

PR #149, opened 2020-11-13
Fix large collection validation

d-w-johnson:
Attempting to serialize a large linked list would result in a stack overflow since our object validation reflectively validates each field in the object being serialized. Sine each node has a reference to the next node it was validating the entire list via recursion in a single stack. Now Iterable and Map objects have special handling to avoid this type of issue by validating each individual item in the collection.


bwhitneyXima (approved):
Looks good.

PR #152, opened 2021-02-22
File exists optimization

d-w-johnson:
Added new FileUtils method for checking if a file for a given path
exists. The Files.exists method is apparently not very performant when
the file doesn't exist. This change should optimize bluedb a bit and
take some load off of the garbage collector. There were tons of
exceptions being thrown, caught, and garbage collected.


athatcher (approved):
Looks great.

PR #154, opened 2021-05-21
XIMA-9485 If there is an extra folder or a folder without a keyType f…

Kaden-Call:
XIMA-9485 If there is an extra folder or a folder without a keyType file in it, when backing up it now skips that folder instead of throwing a nullPointer and breaking the backup.


BlueDBOnDisk/src/main/java/org/bluedb/disk/ReadWriteDbOnDisk.java
```diff
@@ -172,6 +172,7 @@ protected static void assertExistingCollectionIsType(ReadableCollectionOnDisk<?>
 	protected List<ReadWriteCollectionOnDisk<?>> getAllCollectionsFromDisk() throws BlueDbException {
 		List<File> subfolders = FileUtils.getSubFolders(path.toFile());
 		List<ReadWriteCollectionOnDisk<?>> collections = Blutils.map(subfolders, (folder) -> getUntypedCollectionForBackup(folder.getName()));
+		collections.removeIf(readWriteCollectionOnDisk -> readWriteCollectionOnDisk == null);
```
d-w-johnson, line 175: Blutils.map can't return null and it can't return a list that contains null. So this line seems unnecessary.

BlueDBOnDisk/src/test/java/org/bluedb/disk/ReadableDbOnDiskTest.java
```diff
@@ -64,6 +64,12 @@ public void test_getUntypedCollectionForBackup() throws Exception {
 		} catch (Throwable e) {
 		}
 	}
+    //This test we want to pass into getUntypedCollectionForBackup a string for a filename which does not contain a .meta key_type file
+    //Currently it is causing a null pointer when this happens
+	@Test
+    public void test_getUntypedCollectionForBackup_withInvalidPath() throws BlueDbException {
+        assertNull(db.getUntypedCollectionForBackup("bad_folder"));
```
d-w-johnson, line 71: So is this just a folder that doesn't exist at all? We should probably create an empty folder as well that does exist and verify we have the expected result.

d-w-johnson:
So you've taken the following approach:
1 - Make sure that if `keyType` is null nothing can break in the `ReadableCollectionOnDisk` constructor
2 - Make sure that after constructing a concretion of `ReadableCollectionOnDisk`, check to see if it is really valid or not.

But we know that a `ReadableCollectionOnDisk` just isn't valid if the `keyType` or `segmentSizeSetting` isn't set correctly. So what if we took this approach instead?
1 - Make the `ReadableCollectionOnDisk#determineKeyType` method throw an exception if it can't figure out a valid `keyType`.
2 - Make the `ReadableCollectionOnDisk#determineSegmentSize` method throw an exception if it can't figure out a valid `segmentSize`.
3 - Don't catch the exception in the `ReadableCollectionOnDisk`.
4 - Update `ReadWriteDbOnDisk#getUntypedCollectionForBackup` to catch the exception, print the stacktrace, but return null so that it'll just move on to the next collection.
5 - Verify that your test hits the line that returns null.

With this strategy we can make sure that whoever calls this will either have an exception be thrown or they will get a valid collection. That means that they won't have to know that they need to do additional checks in case specific data didn't get set right.

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/ReadWriteCollectionOnDisk.java
```diff
@@ -61,8 +61,8 @@ public ReadWriteCollectionOnDisk(ReadWriteDbOnDisk db, String name, Class<? exte
 		fileManager = new ReadWriteFileManager(serializer);
 		recoveryManager = new RecoveryManager<T>(this, getFileManager(), getSerializer());
 		Rollupable rollupable = this;
-		segmentManager = new ReadWriteSegmentManager<T>(collectionPath, fileManager, rollupable, segmentSizeSettings.getConfig());
 		indexManager = new ReadWriteIndexManager<T>(this, collectionPath);
+		segmentManager = new ReadWriteSegmentManager<T>(collectionPath, fileManager, rollupable, segmentSizeSettings.getConfig());
```
d-w-johnson, line 65: You changed the order of these lines, not sure if that actually matters.

BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/ReadableCollectionOnDisk.java
```diff
@@ -140,6 +140,11 @@ protected static SegmentSizeSetting determineSegmentSize(ReadableCollectionMetad
 
 	protected static Class<? extends BlueKey> determineKeyType(ReadableCollectionMetadata metaData, Class<? extends BlueKey> providedKeyType) throws BlueDbException {
 		Class<? extends BlueKey> storedKeyType = metaData.getKeyType();
+		
+		if(storedKeyType == null && providedKeyType == null) {
+			throw new BlueDbException("Unable to determine key type for collection at " + metaData.getPath() + ". You must initialize a pre-existing valid collection or provide a key type.");
+		} 
+
```
d-w-johnson, line 147: This takes care of causing the constructor to throw an exception if the collection is invalid due to not being able to determine a key type.

BlueDBOnDisk/src/test/java/org/bluedb/disk/ReadableDbOnDiskTest.java
```diff
@@ -754,6 +765,25 @@ public void test_getAllCollectionsFromDisk() throws Exception {
...
+	@Test
+    public void test_getAllCollectionsFromDiskWithNullKeyType() throws Exception {
+        try {
+	        db().getCollectionBuilder("null_key", null, String.class).build();
+	        fail("Building a new collection with a null key type should have thrown an exception");
+        } catch(BlueDbException e) {
+        	//This is what we expect
+        }
+		
+        ReadWriteDbOnDisk newDatabase = (ReadWriteDbOnDisk) new BlueDbOnDiskBuilder().withPath(dbPath).build();
```
d-w-johnson, line 777: BlueDB caches references to the collections, so you have to build a new bluedb instance to get it to get it to actually initialize them from disk.

PR #157, opened 2021-07-13
Skip object validation on individual bytes in a byte array. Also, sets source compatibility to java 8.

d-w-johnson:


BlueDBOnDisk/src/main/java/org/bluedb/disk/serialization/validation/ObjectValidation.java
```diff
@@ -144,4 +145,11 @@ private static boolean isNotEmptyCollection(Object obj) {
 		}
 		return true;
 	}
+
+	private static boolean isByteArray(Field field) {
+		if(field.getType().equals(byte[].class) || field.getType().equals(Byte[].class)) {
```
TheSirwaffle, line 150: Super small thing but you could just return the result of this conditional. If you think it looks more readable without doing that then you don't need to change it.

NathanGift, line 150: I'm fine with leaving it if you'd like, but could this method be simplified to:
 `return field.getType().equals(byte[].class) || field.getType().equals(Byte[].class);`

d-w-johnson: fixed

d-w-johnson: fixed

NathanGift (approved):

PR #159, opened 2021-09-02
Xima 9704

athatcher:
We were experiencing issues where the recovery folders were growing too large and loading the contents of the entire folder resulted in OOM errors. We now use a `DirectoryStream` to reduce memory usage

### DIFF
--- a/BlueDBOnDisk/src/main/java/org/bluedb/disk/file/FileUtils.java
+++ b/BlueDBOnDisk/src/main/java/org/bluedb/disk/file/FileUtils.java
@@ -11,12 +11,15 @@ import java.io.FileOutputStream;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.file.DirectoryStream;
+import java.nio.file.DirectoryStream.Filter;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.LinkedList;
 import java.util.List;
 
 import org.bluedb.api.exceptions.BlueDbException;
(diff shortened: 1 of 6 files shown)


BlueDBOnDisk/src/main/java/org/bluedb/disk/file/FileUtils.java
```diff
@@ -45,10 +48,25 @@ protected FileUtils() {}  // just to get test coverage to 100%
 		return getFolderContentsExcludingTempFiles(path.toFile(), endsWithSuffix);
 	}
 
+	public static DirectoryStream<Path> getFolderContentsExcludingTempFilesAsStream(Path path, String suffix) throws IOException {
+		FileFilter endsWithSuffix = (f) -> f.toPath().toString().endsWith(suffix);
+		return getFolderContentsExcludingTempFilesAsStream(path.toFile(), endsWithSuffix);
+	}
+
+	private static DirectoryStream<Path> getFolderContentsExcludingTempFilesAsStream(File folder, FileFilter filter) throws IOException {
```
d-w-johnson, line 56: I'd probably put this method after the next one so that it reads more like a newspaper. Both methods that use it are listed before, then you see the details below both of them. Not a big deal though and maybe not everyone's preference.

BlueDBOnDisk/src/main/java/org/bluedb/disk/file/FileUtils.java
```diff
@@ -45,10 +48,25 @@ protected FileUtils() {}  // just to get test coverage to 100%
 		return getFolderContentsExcludingTempFiles(path.toFile(), endsWithSuffix);
 	}
 
+	public static DirectoryStream<Path> getFolderContentsExcludingTempFilesAsStream(Path path, String suffix) throws IOException {
+		FileFilter endsWithSuffix = (f) -> f.toPath().toString().endsWith(suffix);
```
d-w-johnson, line 52: Shouldn't this filter use the IS_NOT_TEMP_FILE filter in addition to the ends with filter in order to exclude the temp files?

d-w-johnson: Oh, nvm. It calls a method that adds the is not temp... sorry.

BlueDBOnDisk/src/main/java/org/bluedb/disk/recovery/RecoveryManager.java
```diff
@@ -108,6 +110,10 @@ public void markChangePending(Path completedPath) throws BlueDbException {
 		return FileUtils.getFolderContentsExcludingTempFiles(historyFolderPath, SUFFIX_COMPLETE);
```
d-w-johnson, line 110: I don't want this method anymore. I don't think we should have any methods for getting pending or completed change files that will have problems if there are too many files in the directory if we can help it.

athatcher: Handled in next commit

BlueDBOnDisk/src/main/java/org/bluedb/disk/file/FileUtils.java
athatcher: Good call. Moved

PR #160, opened 2021-10-19
Major BlueDB Improvements

d-w-johnson:
- Made write queries that affect a large number of records avoid loading them all into memory at once.
- Standardized how write queries are executed and how pending changes work which makes the code cleaner and sets us up to potentially execute multiple write queries at the same time in the future.
- You can now provide a key value pair iterator when calling batchUpsert, which makes it possible upsert a large amount of data without holding all of that data in memory at one time. Doing a single pass during an upsert can be much more efficient. Just be aware that only one write can be executed at a time per collection, so executing a huge write will block other writes until it completes. Therefore, it is still probably worth breaking up massive upserts.
- Fixed a problem where batchUpsert didn't update index files appropriately for updated records.
- Added a "whereKeyIsIn" clause to queries which gives users a much better way to match a bunch of records by key. And it allows BlueDB to execute the query in a much more efficient way. For example, if a key from last year and a key from today are requested then we don't need to search the whole year in between the two records.
- The batchDelete method has been removed. Instead, you should use a regular delete query and utilize the "whereKeyIsIn" method.
- Added a way to create multiple indices at once, which can be much more efficient. If a collection has more than one index it is recommended to create them all in a single command rather than one at a time.
- If the process is stopped before index files are completed it will finish creating them on startup.
- Addressed some minor backup bugs.
- Added gradlew to the project.


athatcher (commented):
OK I have reviewed all of the changes and things look great. There is one variable name that I think is misleading. Have a look at that and then I think this is ready to merge.

PR #162, opened 2022-04-28
Index Conditions, Version 2 Time Collections, and other optimizations.

d-w-johnson:
- Index files are kept up to date a little better now.
-  Collection queries now support index conditions in a where clause. This allows you to utilize indexed data in any type of query in order to optimize that query.
- BlueIndex#get(key) has been removed. You can use index conditions on a query instead, which are much more powerful.
- Removed BlueDB#initializeCollection and BlueDB#collectionBuilder methods which had already been deprecated in favor of newer methods.
- Optimized whereKeyIsIn and index condition queries. They know exactly what keys match the query, so we can avoid searching the whole timeframe now that we utilize that information better.
- The IndividualChange class is registered with FST now which should save disk space and resources on large queries. Since we persist the changes before executing write updates for recovery purposes.
- Time records will roll up to 6 seconds then 5 minutes, then 1 hour now. The added 5 minutes rollup level should be useful because it avoids accidentally rolling up the whole segment sooner than we want to in cases where we batch a bunch of updates that cover more than 6 seconds. Prematurely rolling up the whole current segment would add a lot of disk usage as frequent changes continue to be made after rolling it up.
- After deletion, empty files and segment directories are cleaned up better now.
- Version 2 collections exist now and can be optionally used when creating a new collection for the first time. Version 2 time collections have the following benefits:
  - They treat values with an active time key as active, meaning that the values only have a start time, but are considered to be active in all timeframes at or after that time.
  - You can easily designate whether to include or exclude active records in a query where clause.
  - A value with a TimeFrameKey will only be stored in one segment. Time indexes are automatically created which allow BlueDB to still identify what active and completed records overlap with a query timeframe without duplicating these records in each segment.
  - All time keys are considered equal as long as the id and time are the same. Two TimeFrameKey records with different end times will be considered the same key. The end time is useful for upkeeping the time index, but is no longer considered part of the value's primary key. So you should include it when inserting a record that has a start and end time, but you don't need to include it when trying to find a value with a given id and time. This makes it much easier to deal with records that could have the end time change or be considered active and then completed.
  - Active and Completed records can be stored in the same collection now. An active record can be completed, and a completed record can have its end time changed without having to delete it and re-add it with a new key. This makes it much easier to deal with this type of data rather than having to add complex, error prone work arounds.
  - You can pass a new but equivalent time key into BlueTimeCollection#updateKeyAndValue or BlueTimeCollection#replaceKeyAndValue or call BlueTimeQuery#updateKeyAndValue or BlueTimeQuery#replaceKeyAndValue in order to update a record in a way that affects its time key. It must be considered the same key, but you can use this to make an active record completed or to change the end time on a completed record without deleting and re-inserting the record.
- Updated the readme to have a better information for devs on how to verify proper test coverage and build artifacts.
- BlueDbSample and QuickStart Java classes have been updated to include examples of what you can do with index conditions now.


BlueDBOnDisk/src/main/java/org/bluedb/disk/collection/index/extractors/OverlappingTimeSegmentsKeyExtractor.java
```diff
@@ -0,0 +1,41 @@
...
+	public Class<LongTimeKey> getType() {
+		return LongTimeKey.class;
+	}
+	
+	@Override
+	public List<LongTimeKey> extractKeys(BlueKey key, ReadableSegmentManager<V> segmentManager) {
+		LinkedList<LongTimeKey> keys = StreamUtils.stream(segmentManager.getPathManager().getAllPossibleSegmentStartGroupingNumbers(key))
+			.map(LongTimeKey::new)
+			.collect(Collectors.toCollection(LinkedList::new));
+		keys.remove(0);
```
athatcher, line 32: Why are we removing the `0` index here? Does that need to happen every time. Is it possible that the stream gives us an empty list?

d-w-johnson: All keys have one grouping number. The method it calls will always at least return that grouping number. If the key overlaps more than one segment then it would also return grouping numbers for those segments. We are removing the first grouping number since we don't care about tracking that in the index. We know how to find a value in the segment that it starts in. We want to know what segments it overlaps, so that when we're querying those segments we can easily find the records that started sometime in the past and overlap into our segment.

That being said, it is irrelevant now. I just pushed a commit that refactors that a little bit and does it in a different way. Sorry about that, I didn't realize I forgot to push that last commit.

athatcher: Ha, well I will look at the new commit. Thanks for the explanation either way

